PR #5, opened 2020-11-01
Bowling Challenge

emilyalice2708:


lib/bowling_game.rb
```diff
@@ -0,0 +1,104 @@
...
+    elsif first_roll?
+      create_new_frame(pins, frame)
+    else
+      close_frame(pins, frame)
+    end
+  end
+
+  def score
+    calculate_score
+    @score
```
tatsiana-makers, line 25: Does `@score` need to be an instance variable which is stored within the Bowling Game class state. Could we instead calculate and return the result to the user without using an instance variable?

Storing it as state causes issues with the current code:
```
>bg = BowlingGame.new
>bg.bowl(3)
> bg.bowl(6)
> bg.score
 => 9
> bg.score
 => 18
```

As you can see in this example (which I copied from irb), by always doing += on the stored state, we'll keep incrementing the score instance variable each time we call `.score`. If we instead calculated and returned the score without storing it, we wouldn't have this problem because after we returned the local variable it would go out of scope and become inaccessible next time we called `.score`.

spec/feature/scoring_games_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    subject.bowl(4)
+  
+    subject.bowl(10)
+  
+    subject.bowl(2)
+    subject.bowl(8)
+    subject.bowl(6)
+    expect(subject.score).to eq 133
+  end
+end
```
tatsiana-makers, line 87: Awesome test coverage of the different scenarios here!

spec/unit/bowling_game_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'bowling_game'
+
+describe BowlingGame do
+  let(:game) { described_class.new }
+  let(:frame) { double :frame }
+
+  describe '#bowl' do
+    it 'can begin a new frame' do
+      allow(frame).to receive(:new).and_return(frame)
```
tatsiana-makers, line 9: I think we might want to return a different double - an double of an instance of the frame class. An example of this can be seen [here](https://github.com/makersacademy/skills-workshops/blob/f5b4801840fe07d26ff70341652dc81dcda12289/practicals/object_oriented_design/dependency_injection.md#intermezzo-of-classes)

spec/unit/bowling_game_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  describe '#score' do
+    it 'can check for strikes and spares' do
+      allow(frame).to receive(:new).and_return(frame)
+      allow(frame).to receive(:total).and_return(10)
+      expect(frame).to receive(:add_to_frame).with(10)
+      expect(frame).to receive(:add_to_frame).with(0)
+      game.bowl(10, frame)
+      expect(frame).to receive(:strike?)
+      expect(frame).to receive(:spare?)
+      game.score
```
tatsiana-makers, line 45: Should we also check that `score` returns the right value here?

tatsiana-makers (commented):
This is a nice go at the Bowling challenge! I found your code readable. You've shown a nice separation of concerns with the two classes and encapsulation with private methods. You've organised your test files in a methodic way which makes it easy to read the tests and review them, nice! I've left a few comments - let me know if you have any questions about them, happy to chat further! 🙂

PR #12, opened 2020-11-29
Weekend challenge(5?) done!

MykeNuLeng:


lib/bowling.rb
```diff
@@ -0,0 +1,78 @@
+class Bowling
+  attr_reader :bonus, :total, :frame, :current_frame
```
alicelieutier, line 2: Why do you need to expose all of these instance variables? Could some stay private?

spec/bowling_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'bowling'
+describe Bowling do
+  it 'it responds to add_roll' do
+    expect(subject).to respond_to :add_roll
+  end
```
alicelieutier, line 5: You can remove this test. As soon as you test `add_roll`, this is redundant. I would typically skip it at your level.

spec/bowling_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    new_game = Bowling.new
+    new_game.add_roll(6)
+    expect(new_game.total).to eq 6
+  end
+
+  it 'scoring a spare applies bonus' do
+    new_game = Bowling.new
+    new_game.add_roll(6)
+    new_game.add_roll(4)
+    expect(new_game.bonus).to eq 1
```
alicelieutier, line 17: Are users going to use `bonus` and `total` directly to know the overall score? Or should they be private? Depending on the answer, this is maybe testing implementation more than behaviour.

You want to first think about the interface (how the user will use the class, which methods are public), and then write you tests using on what the user could access. This is so that your tests are not locking you into a specific implementation, or specific vraiable names and type. The only thing fixed is the interface.

spec/bowling_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    new_game = Bowling.new
+    12.times { new_game.add_roll(10) }
+    expect(new_game.total).to eq 300
+  end
+
+  it 'cant roll an extra strike after the perfect game' do
+    new_game = Bowling.new
+    12.times { new_game.add_roll(10) }
+    expect { new_game.add_roll(10) }.to raise_error 'The game is over!'
+  end
```
alicelieutier, line 87: You may want to regroup the endgame tests (all the ones that allow or reject adding a new roll when the game reaches the end, to include strike on 9th frame, strike on 10th frame, spare on tenth frame, and normal rolls

lib/bowling.rb
```diff
@@ -0,0 +1,78 @@
...
+    @frame += 1
+  end
+
+  def multiplier(pins)
+    double_bonus? ? pins * 3 : pins * 2
+  end
+
+  def apply_score(pins, score)
+    if double_bonus?
+      @bonus += -2
```
alicelieutier, line 35: rewrite `@bonus -= 2`

lib/bowling.rb
```diff
@@ -0,0 +1,78 @@
...
+    elsif bonus?
+      @bonus += -1
```
alicelieutier, line 37: rewrite `@bonus -= 1`

lib/bowling.rb
```diff
@@ -0,0 +1,78 @@
...
+    @current_frame.length == 2 && current_frame_sum == 10
+  end
+
+  def strike?
+    @current_frame.length == 1 && @current_frame[0] == 10
+  end
+
+  def bonus?
+    @bonus.positive?
+  end
```
alicelieutier, line 61: This method is not replacing much. Maybe it doesn't need to be extracted.

spec/bowling_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    expect(new_game.total).to eq 18
+  end
+
+  it 'a spares bonus only lasts for 1 round' do
+    new_game = Bowling.new
+    new_game.add_roll(6)
+    new_game.add_roll(4)
+    new_game.add_roll(4)
+    expect(new_game.bonus).to eq 0
+  end
```
alicelieutier, line 34: Maybe add a test that mixes strike and spare bonuses. (Spare then strike, and strike then spare)

lib/bowling.rb
```diff
@@ -0,0 +1,78 @@
...
+  def game_continues?
+    @frame < 10 || (@frame == 10 && tenth_frame_exception?)
+  end
+
+  def tenth_frame_exception?
+    @current_frame.length < 2 || (current_frame_sum >= 10 && @current_frame.length < 3)
+  end
+
+  def current_frame_sum
+    @current_frame.inject(0, &:+)
```
alicelieutier, line 76: Rewrite `@current_frame.sum`

alicelieutier: Maybe even no need to extract as it's a one line with only one operation.

alicelieutier (commented):
Complex logic, but your code is well organised and overall well tested.
You may want to refine your class interface slightly. Next week will be the same challenge in JS, which can be an opportunity to refactor even further (maybe separate classes?)

lib/bowling.rb
MykeNuLeng: Completely valid, only total and perhaps frame and current frame would be relevant to the user!

lib/bowling.rb
MykeNuLeng: This really is a bad habit that I picked up in the pre-course, I'll work on that

lib/bowling.rb
MykeNuLeng: I definitely went overboard with extracting classes.

spec/bowling_spec.rb
MykeNuLeng: Thanks, this is another habit that I've gotten into, should be fairly easy to stop doing that.

spec/bowling_spec.rb
MykeNuLeng: I was definitely planning on using total to calculate the scores, but bonus is definitely something that they shouldn't have access to. It's odd, I've heard coaches on this course say 'dont test implementation, test the behaviour' but this is the first time that it's clicked.

spec/bowling_spec.rb
MykeNuLeng: That's a good idea

spec/bowling_spec.rb
MykeNuLeng: Oooh, I like that a lot.

lib/bowling.rb
MykeNuLeng: methods, I meant methods...

PR #14, opened 2020-11-30
Bowling Challenge - Partial Solution

Ephyrean:


lib/frame.rb
```diff
@@ -0,0 +1,56 @@
...
+    @state = state
+  end
+
+  def first_roll(pins)
+    return strike if is_strike?(pins)
+    increment_score(:first_roll, pins)
+  end
+
+  def second_roll(pins)
+    @@spare_status = true if is_spare?(pins)
```
joshuaretallick, line 27: This is such a cool way to think about the spare concept!

spec/game_spec.rb
```diff
@@ -0,0 +1,3 @@
+describe Game do
+
+end
```
joshuaretallick, line 3: I cant claim credit for any of the stuff in my code but I think the testing approach was smart and may have been useful for yourself too.  Eg 'it scores all 0's', or 'it scores all 1's', 'it scores a strike'
Maybe asking ourselves these questions in our tests can actually help us with what code we should write.

joshuaretallick (commented):
Really nice looking code Constantine, I hope you get some time to finish this!

PR #15, opened 2020-11-30
Partial solution

alexleesonmill:
Completed everything apart from stopping a player rolling more than 10 in one frame. 

Probably could have eventually moved it all into a separate class but I ran out of time and felt like I’d forgotten all ruby so didn’t manage

### DIFF
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 
 class Game
-  attr_reader :strike_count, :frame, :bonus
+  attr_accessor :score, :frame
 
+  TP = 10
+  TF = 10
   def initialize
     @score = []
     @frame = 0
-    @bonus = 0
   end
 
   def roll(pins)
@@ -19,19 +20,17 @@ class Game
     result = 0
     index = 0
     10.times do
+      @frame += 1
       if strike?(index)
-        result += strike_score(index)
+        result += last_frame ? bonus(index) : strike_score(index)
         index += 1
-        # strike_bonus if @frame >= 10 && @frame <= 12
       elsif spare?(index)
-        result += spare_score(index)
+        result += last_frame ? bonus(index) : spare_score(index)
         index += 2
-        # spare_bonus if @frame >= 10 && @frame <= 11
       else
         result += frame_score(index)
         index += 2
       end
-      @frame += 1
     end
     result
   end
@@ -39,19 +38,19 @@ class Game
   private
 
   def game_over?
-    @score.count >= 20 || @frame >= 10
+    @score.count >= 21 || @frame >= TF
   end
 
   def strike?(index)
-    @score[index] == 10
+    @score[index] == TP
   end
 
   def spare?(index)
-    @score[index] + @score[index + 1] == 10
+    @score[index] + @score[index + 1] == TP
   end
 
   def strike_score(index)
-    10 + @score[index + 1] + @score[index + 2]
+    TP + @score[index + 1] + @score[index + 2]
   end
 
   def spare_score(index)
@@ -62,13 +61,16 @@ class Game
     @score[index] + @score[index + 1]
   end
 
-  # def strike_bonus
-  #   10 + @score[index] + @score[index + 1]
-  #   @bonus += 1
-  # end
+  def bonus(index)
+    if strike?(index)
+      @frame -= 2
+      strike_score(index)
+    else
+      spare_score(index)
+    end
+  end
 
-  # def spare_bonus
-  #   10 + @score[index]
-  #   @bonus += 1
-  # end
+  def last_frame(frame = @frame)
+    frame == TF
+  end
 end
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Game
   def initialize
     @score = []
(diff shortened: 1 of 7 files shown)


lib/game.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require 'frame'
+
+class Game
+  attr_reader :frame, :scores
+
+  TP = 10
+  TF = 10
```
AdamusBG, line 9: Why are these two constants separate but of the same value?

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+    @scores[index] + @scores[index + 1] + @scores[index + 2]
+  end
+
+  def frame_score(index)
+    @scores[index] + @scores[index + 1]
+  end
+
+  def bonus(index)
+    if strike?(index)
+      @frame -= 2
```
AdamusBG, line 68: I see, this lets you continue when there's a strike or a spare nice!

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def strike?(index)
+    @scores[index] == TP
+  end
+
+  def spare?(index)
+    @scores[index] + @scores[index + 1] == TP
+  end
+
+  def strike_score(index)
```
AdamusBG, line 54: This plus next two methods looking good!

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+        index += 2
+      end
+    end
+    result
+  end
+
+  private
+
+  def game_over?
+    @scores.count >= 21 || @frame >= TF
```
AdamusBG, line 43: nice, stopping the game from finishing here!

spec/game_spec.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+describe 'A game of bowling' do
```
AdamusBG, line 3: Nice, thoroughly tested

AdamusBG (commented):
Nice, well tested! Could consider splitting into one more class maybe.

lib/game.rb
alicelieutier: It would be easier if the constants had explicit names - using abbreviations are not recommended as it make it difficult to know what variables contain.

See this article for more on naming: https://medium.com/swlh/an-opinionated-guide-to-naming-your-code-aimed-at-new-developers-6f769d60e098

spec/game_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  let(:game) { Game.new }
+  let(:rolls) { 17.times { game.roll(0) } }
```
alicelieutier, line 5: Given that you are only replacing one line by one line here, it probably would make it more readable to just have the rolls in the tests.

spec/game_spec.rb
alicelieutier: Agreed, this are really good tests, focusing on behaviour.

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+    @frame = 0
+  end
+
+  def roll(pins)
+    raise 'Your game is now complete' if game_over?
+
+    @scores << pins
+  end
+
+  def score
```
alicelieutier, line 21: Can this method only be called when the game is over? This is the only way you use it in your tests.
If that is the case, maybe it should give an error if called before the end of the game?
If not, you may want to test partial scores.

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+    result = 0
+    index = 0
+    TF.times do
+      @frame += 1
```
alicelieutier, line 25: Can you use a local variable to iterate over frames here?

alicelieutier: More generally, instance variables should be storing things that say something about the instance, and need to live for as long as the instance lives. They are not just shared memory between methods. If you want to pass state to a different method called by a first method, you should use an argument.

spec/game_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  it 'stops after the bonus rolls' do
+    rolls
+    game.roll(0)
+    game.roll(10)
+    game.roll(10)
+    game.roll(10)
+    expect { game.roll(10) }.to raise_error('Your game is now complete')
+    expect(game.score).to eq 30
+  end
```
alicelieutier, line 77: It would be nice to test bonus rolls that are not themselves strikes.

lib/game.rb
alicelieutier: Given that you don't allow rolls to continue after game over, you don't really need this method.

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+      if strike?(index)
+        result += last_frame ? bonus(index) : strike_score(index)
```
alicelieutier, line 27: What would happen if this was just: `result += strike_score(index)` ?

spec/game_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  it 'stops bonus frame if game doesnt end with spare or strike' do
+    9.times { game.roll(10) }
+    game.roll(5)
+    game.roll(4)
+    expect(game.score).to eq 263
```
alicelieutier, line 83: Sneaky.
You have to call score for your game over to work here, which is not ideal.

lib/game.rb
alicelieutier: This method should purely work based on the score array, as the frame variable is not updated until you call the score.
Think about how you would know whether a game has finished or not looking simply at the score array. Is there a way?

alicelieutier (commented):
Overall your tests are good and your code is pretty clean. There are a few design flows you can improve - maybe use the javascript version next weekend for that!

PR #19, opened 2020-12-01
bowling-challenge-ruby

sedwards93:


spec/frame_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  describe '#frame_check' do
+    it 'Calls the close_play method if 2 throws have happened' do
+      expect(frame).to receive(:close_play)
+      2.times {frame.knocked(2)}
+    end
+  end
+
+  describe '#strike_check' do
+  it 'Calls the close_play method if 2 throws have happened' do
```
tatsiana-makers, line 30: minor note: I think this should read "if a strike has happened"

spec/game_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    end
+  end
+
+  describe '#score' do
+    it 'Takes into account strike bonus' do
+      game = Game.new
+      game.roll(10)
+      game.roll(3)
+      game.roll(4)
+      expect(game.frames[0].score).to eq 17
```
tatsiana-makers, line 35: Should we also assert on the total score of the game so far? We could do this across the tests

spec/game_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+      game.roll(5)
+      game.roll(4)
+      game.roll(4)
+      game.roll(4)
+      game.roll(4)
+      game.roll(10)
+      game.roll(1)
+      game.roll(1)
+      expect(game.frames[3].score).to eq 12
+    end
```
tatsiana-makers, line 65: Are these tests necessary? 

It seems like they test the same scenario as the strike test [here](https://github.com/makersacademy/bowling-challenge-ruby/pull/19/files#diff-3f3416f9e912511dc1f61fc61d2832a11dae55401e24c04209cd84f7c40ad633R30-R36)

spec/game_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    it 'Takes into account spare bonus' do
+      game = Game.new
+      game.roll(5)
+      game.roll(5)
+      game.roll(4)
+      game.roll(3)
+      expect(game.frames[0].score).to eq 14
+      end
+      it 'Takes into account spare bonus' do
```
tatsiana-makers, line 74: Similarly, this test seems to cover the same thing as the test above so can be removed

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+    else frame_count == 10 && current_frame.in_play
+         current_frame.knocked(pins)
+    end
+  end
+
+  def total_score
+    total_score = @frames.map {|frame | frame.score }.sum
+  end
+
+  def last_frame_strike?
```
tatsiana-makers, line 38: It seems like this method could be made private because we're not using it outside of this class and it's not a method that the end-user would need to call themselves. Are there other methods in both this and the Frame class which could be made private?

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+  attr_reader :frames
+  
+  def initialize
+    @frames = [Frame.new]
+  end
+
+  def roll(pins)
+    case 
+    when frame_count == 1 && current_frame.in_play
+         current_frame.knocked(pins)
```
tatsiana-makers, line 13: Could we combine this case with the `frame_count <=8 && current_frame.in_play` scenario?
This would mean amending `last_frame_strike?` or `last_frame` to handle when the `frame_count` is 1, but would mean one fewer case

spec/game_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    end
+  end
+
+  context 'Perfect Game' do
+    it 'Scores 300 across 12 rolls' do
+      12.times {game.roll(10)}
+      expect(game.total_score).to eq 300
+    end
+  end
+
```
tatsiana-makers, line 115: There could also be a test for when there are no strikes/spares/gutter balls

tatsiana-makers (commented):
Thanks for asking for a review! This is a nice go at the bowling challenge and I found your logic clear to follow overall 🙂  I've left some suggestions. Let me know if you have any questions - happy to chat further

lib/game.rb
```diff
@@ -0,0 +1,72 @@
+require_relative 'frame'
+
+class Game
+  attr_reader :frames
+  
+  def initialize
```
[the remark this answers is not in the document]
sedwards93: Hi @tatsiana-makers. I feel like I had cracked dependency injection but now I seem to be lost again... Because the two classes were so tightly coupled, as soon as I began to mock the frame double methods, the tests seem become kind of redundant?  😖 Before I could check that a series of specific rolls were tallying up correctly, but now they don't? Am I missing something. Thank you!

PR #32, opened 2021-01-11
Bowling challenge

ghost:


WillDixon93:
can't get the strike and strike bojnus to work, seems to think its NaN instead of an integer.

src/bowling.js
```diff
@@ -0,0 +1,49 @@
...
+
+  constructor(){
+    this.rolls = [];
+  }
+
+  roll(pins) {
+    this.rolls.push(pins)
+  }
+
+  get score() {
```
chrismabdo, line 11: Good use of a get method, haven't used them myself seems to have be useful here

src/bowling.js
```diff
@@ -0,0 +1,49 @@
...
+    let score = 0;
+    let rollIndex = 0;
+
+    for (let frameIndex = 0; frameIndex < 10; frameIndex++) {
```
chrismabdo, line 15: Nice use of a loop to keep turns inclusive of rounds 1-10

src/bowling.js
```diff
@@ -0,0 +1,49 @@
...
+      
+      if (this.isStrike(rollIndex)) {
+        score += this.strikeBonus(rollIndex);
+        rollIndex++;
+        continue;
```
chrismabdo, line 20: Haven't seen the 'continue;' keyword before. Is there a reason why its needed here but not in other places?

spec/bowlSpec.js
```diff
@@ -0,0 +1,44 @@
...
+   game.roll(1);
+   expect(game.score).toEqual(14)
+ });
+
+//  it('should return 300 for a perfect game', function(){
+//    rollMany(12, 10)
+//    expect(game.score).toEqual(300);
+//  })
+
+ function rollMany(pins, rolls) {
```
chrismabdo, line 39: Nice use of a function here, keeps tests thorough but clean

spec/bowlSpec.js
```diff
@@ -0,0 +1,44 @@
...
+  for (let i = 0; i < 20; i++) {
+    game.roll(pins);
+ }
+}
+});
```
chrismabdo, line 44: Perhaps could add in a few more tests such as consecutive strikes/consecutive spares, test against edge cases etc.

PR #38, opened 2021-01-20
Bowling (Ruby)

PiperS52:


README.md
```diff
@@ -56,10 +30,31 @@ A Gutter Game is when the player never hits a pin (20 zero scores).
...
+```
+scorecard = ScoreCard.new
+scorecard.run
+```
+### Completed
+- Scores are correct for ten frames (inc bonus)
+- Programs runs as intended
+
+### Queries
+- Any guidance on how to implement further feature tests, as requiring user input to generate scores seems to make this slightly more fiddly (?)
```
tatsiana-makers, line 58: This [answer](https://stackoverflow.com/a/46515819) shows a way of simulating multiple user inputs. Does this help when considering how to make feature tests?

README.md
```diff
@@ -56,10 +30,31 @@ A Gutter Game is when the player never hits a pin (20 zero scores).
...
+- Please could you check correct use of stubs
+- Originally planned to use one or two other classes, although with the approach I took, running it in the terminal, I ended up keeping it to a single class
```
tatsiana-makers, line 60: In the scorecard code, there is a class that we could extract out. Currently we store the scores for each frame in an array. This indicates that we could potentially extract out a Frame class to encapsulate the scores for that frame

spec/scorecard_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+  describe '#roll_1_input' do
+    it "should take a user input for their first roll" do
+      expect(scorecard).to receive(:gets).and_return("5\n")
+      scorecard.roll_1_input
+    end
+
+    it "should add the inputted score for roll 1 to frame_score" do
+      allow(scorecard).to receive(:gets).and_return("5\n")
+      scorecard.roll_1_input
+      expect(scorecard.frame_score).to eq([5])
```
tatsiana-makers, line 23: These tests pass which suggest that the stubbing is working as intended 🙌

lib/scorecard.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def run
+    while @frames < 10 do
+      roll_1_input
+      if strike?
+        @frames += 1
+        clear_frame
+      else
+        roll_2_input
+        @frames += 1
```
tatsiana-makers, line 21: We do
```
@frames += 1
clear_frame
``` 
to the frame in both the if and the else branch, so we could extract this out of the if/else and refactor to look like:
```
if !strike?
    roll_2_input
end
@frames += 1
clear_frame
```
With this format, it's easier for a reader of the code to see that there is logic that happens whether there's a strike or not and DRYs out the code

tatsiana-makers: We could do a similar refactor for the 10th roll logic

lib/scorecard.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def bonus_roll_input
+    puts "Please enter roll 3 score"
+    bonus_roll = gets.chomp.to_i
+    @scores << bonus_roll
+    @frame_score << bonus_roll
+  end
+
+  def strike?
+    @frame_score[0] == 10
```
tatsiana-makers, line 63: 10 is a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) that appears throughout the code. We could refactor to have a clearly-named constant set to 10 to make the program more readable

lib/scorecard.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def spare?
+    (@frame_score[0] + @frame_score[1]) == 10
+  end
+
+  def clear_frame
```
tatsiana-makers, line 70: Can any of the methods in this class be private?

tatsiana-makers (commented):
Thanks for asking for a review! I've left a few comments, let me know if you want to chat about any of them further 🙂

spec/scorecard_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    expect(scorecard).to respond_to(:run).with(0).arguments
+  end
+
+  describe '#roll_1_input' do
+    it "should take a user input for their first roll" do
+      expect(scorecard).to receive(:gets).and_return("5\n")
+      scorecard.roll_1_input
+    end
+
+    it "should add the inputted score for roll 1 to frame_score" do
```
tatsiana-makers, line 20: A few of these tests are testing state rather than behaviour. [This ](https://github.com/makersacademy/skills-workshops/blob/master/practicals/testing/behaviour_not_state.md)is a practical about this topic 🙂

PR #41, opened 2021-02-07
Bowling Scorecard Challenge (Ruby) - Reeshul

Reeshul:
Working but need to extract out a frame class


JosephSpraggon (commented):
Nice work Reeshul! some very good stuff. Some bugs seem like they need to be ironed out & could maybe do with using some private methods for your logic heavy methods. Other than that really good!

hacaravan (commented):
Nice, the code is really DRY! Not quite sure why it wasn't working in your irb, maybe just the format of the input?

PR #46, opened 2021-02-08
Karsten Finlay - Bowling-challenge-ruby

KarstenFinlay:


lib/score.rb
```diff
@@ -0,0 +1,57 @@
...
+    @scores[-1] = 10 + frame.sum
+  end
+
+  def update_spare(pins)
+    @scores[-1] = 10 + pins
+  end
+
+  def update_consec_strikes(pins)
+    @scores[-2] = 20 + pins
+  end
```
twigz826, line 47: I like the logic here and think I understand what this code is doing.

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  MAX_FRAME = 10
+
+  def initialize(score = Score.new)
+    @throws = []
+    @frame_complete = true
+    @score = score
+    @frame_count = 1
+    @last_frame = []
+  end
```
twigz826, line 14: I'd be interested to play this in irb as I can't quite follow how score.rb and game.rb link up

twigz826 (commented):
This looks great, lots of clearly defined methods and SRP maintained throughout. Would be interested to see how it works in IRB.

lib/game.rb
```diff
@@ -0,0 +1,67 @@
+require_relative './score'
+
+class Game
+  attr_reader :throws, :score, :frame_count, :last_frame
+
+  MAX_FRAME = 10
+
+  def initialize(score = Score.new)
+    @throws = []
+    @frame_complete = true
```
mscwilson, line 10: Why does @frame_complete start on true?

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+      @score.update_if_needed(pins)
+    end
+  end
+
+  def new_frame
+    @throws.clear
+    @frame_count += 1
+  end
+
+  def last_frame?
```
mscwilson, line 45: This method is called the same thing as the @last_frame - makes it seem like a getter method but it isn't. Perhaps one of them could be renamed?

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def add(pins)
+    @throws << pins
+    play_last_frame(pins) if last_frame?
+    frame_status(pins)
+    setup_next_roll(pins)
+  end
+
+  def frame_status(pins)
+    @frame_complete = !@frame_complete if pins != 10
```
mscwilson, line 24: a bit opaque! maybe use the `frame_complete?` method since you have it? That would show that booleans are involved

spec/game_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    it 'calls for a new frame if frame complete' do
+      expect(subject).to receive(:new_frame)
+      subject.add(10)
+    end
+  end
+
+  describe '#last_frame?' do
+    it 'returns true if frame count is 10' do
+      (Game::MAX_FRAME * 2 - 2).times { subject.add(3) }
```
mscwilson, line 62: could make a new constant like THROWS_UNTIL_LAST_FRAME to avoid these 2s

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+    @frame_count >= MAX_FRAME
+  end
+
+  def play_last_frame(pins)
+    @last_frame << pins
+    finished if last_frame_complete?
+  end
+
+  def last_frame_complete?
+    @last_frame.size == 2 && @last_frame.sum < 10 || @last_frame.size == 3
```
mscwilson, line 55: can you add some brackets? at first glance I wasn't sure which bit went with which

lib/score.rb
```diff
@@ -0,0 +1,57 @@
+class Score
```
mscwilson, line 1: nice separation of concerns

lib/game.rb
```diff
@@ -0,0 +1,67 @@
+require_relative './score'
+
+class Game
+  attr_reader :throws, :score, :frame_count, :last_frame
+
+  MAX_FRAME = 10
+
+  def initialize(score = Score.new)
```
mscwilson, line 8: neat way to import the Score methods

lib/score.rb
```diff
@@ -0,0 +1,57 @@
...
+  attr_reader :scores
+
+  def initialize
+    @scores = []
+  end
+
+  def calculate(frame)
+    update_if_needed(frame)
+    if frame == [10]
+      @scores << 'X'
```
mscwilson, line 11: You are using X and /, very nice!

mscwilson (commented):
Nice separation of concerns with the Game/Score classes. Overall quite clean but maybe change a couple of the variable names

lib/game.rb
KarstenFinlay: There really is no reason! Will change that!

PR #48, opened 2021-02-08
Bowling challenge Ruby

jshields123:


README.md
```diff
@@ -1,3 +1,13 @@
+Approach - 
+
+Attempted to create a frame and game class for this and spent a few hours trying to get the planning down.
+
+Instead went with just the one class and developed the tests first for each specification of the game. 
+
+Used this video as found when searching for help. https://www.youtube.com/watch?v=wrr16PdgHPM
```
mscwilson, line 7: nicely honest!

spec/game_spec.rb
```diff
@@ -0,0 +1,49 @@
+require './game'
+
+describe Game do
+  let(:game) { Game.new }
+
+  # perfect game
+  it 'can roll all strikes' do
```
mscwilson, line 7: I suppose this is more of a BDD approach than TDD? Nice

spec/game_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    12.times { game.roll(10) }
+    expect(game.scorecard).to eq 300
+  end
+
+  # guttter game
+  it 'can roll all 0s ' do
+    20.times { game.roll(0) }
```
mscwilson, line 14: Would be good to take the magic numbers out into constants

game.rb
```diff
@@ -0,0 +1,52 @@
...
+    @rolls = []
+  end
+
+  def roll(num_pins)
+    @rolls << num_pins
+  end
+
+  def scorecard
+    result = 0
+    roll_count = 0
```
mscwilson, line 15: I found this roll_count a bit confusing - maybe could be renamed frame_count?

game.rb
```diff
@@ -0,0 +1,52 @@
...
+    result
+  end
+
+  def strike?(roll_count)
+    @rolls[roll_count] == 10
+  end
+
+  def strike_bonus(roll_count)
+    strike = 10
+    strike + @rolls[roll_count + 1] + @rolls[roll_count + 2]
```
mscwilson, line 37: This works because roll count is natural counting numbers whereas the `@rolls[]` is starting from 0? Ah ok I get it, the `@rolls` list is fully populated before scorecard looks at it

game.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def spare?(roll_count)
+    @rolls[roll_count] + @rolls[roll_count + 1] == 10
+  end
+
+  def spare_bonus(roll_count)
```
mscwilson, line 44: nit picking but this is more doing "spare_and_bonus" rather than "bonus"? Since it includes the normal 10 points for the spare

mscwilson (commented):
I did find it a bit difficult to follow, I think partly because I was thinking in terms of frames for mine but yours is in terms of individual rolls. So I wasn't sure what the `roll_count += 2` was about to start with. But it obviously works so fair enough!

PR #49, opened 2021-02-08
Anna's bowling challenge

acavalla:


lib/bonus.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Bonus
+  attr_reader :pins
+  def initialize(num, pins)
+    @pins = pins
+  end
+
+end
```
louiswicks, line 9: What is the Bonus class supposed to do?

louiswicks: Sorry just finished your README and see you say you haven't finished the bonus class completely yet.

lib/frame.rb
```diff
@@ -0,0 +1,40 @@
...
+    @open = false if close?
+  end
+
+  private
+
+  def close?
+    return true if number < 10 && rolls.length == 2
+    return true if number < 10 && strike?
+    return true if rolls.length == 3
+    return true if number == 10 && rolls.length == 2 && !(strike? || spare?)
```
louiswicks, line 26: Very clever close? method, something I didn't even think to do. Nice job.

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    frames << Frame.new(frames.length + 1) if add_frame?
+
+    frames[-1].add_roll(pins)
+  end
+
+  def calculate_score
+    @running_total = 0
+    @score_list.each { |x| @running_total += x[:pins] }
+  end
```
louiswicks, line 28: Great way of calculating final score and still retaining the score information for each frame.

lib/player.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Player
+  attr_reader :bowl
+
+  def roll
+    'What is your roll?'
+    @bowl = $stdin.gets.chomp.to_i
+  end
+end
```
louiswicks, line 10: Is a player class required? Does a player usually have to input their score themselves?

louiswicks (commented):
Your code looks great Anna, I think there are a lot of things you've done I could have used and improved my own code. I think I might need to break my classes up into more than one game class.

lib/player.rb
acavalla: This was an earlier strategy... should definitely delete player haha

PR #51, opened 2021-02-08
Pete's bowling challenge

PetarSimonovic:
It works, but needs lots of refactoring


coverage/.resultset.json
```diff
@@ -0,0 +1,352 @@
...
+          1,
+          1,
+          null,
+          null
+        ]
+      }
+    },
+    "timestamp": 1612775036
+  }
+}
```
HamishArro, line 352: maybe add to .gitignore?

PR #52, opened 2021-02-08
Partial solution - Needs refactoring

OscarB89:


lib/game.rb
```diff
@@ -0,0 +1,40 @@
+class Game
+  def initialize
+    @rolls = []
+  end
+
```
ddottyler, line 5: Does this class do too much? could it be split up?

lib/game.rb
```diff
@@ -0,0 +1,40 @@
...
+        result += spare_result(index)
+        index += 2
+      elsif result += regular_result(index)
+        index += 2
+      end
+    end
+    result
+  end
+
+  private
```
ddottyler, line 27: why are these methods private?

lib/game.rb
```diff
@@ -0,0 +1,40 @@
...
+
```
ddottyler, line 28: Not sure I fully understand the naming of each of these methods. Could it be sign posted any better?

lib/game.rb
```diff
@@ -0,0 +1,40 @@
...
+  def spare(index)
+    @rolls[index].to_i + @rolls[index + 1].to_i == 10
+  end
+
+  def spare_result(index)
+    @rolls[index].to_i + @rolls[index + 1].to_i + @rolls[index + 2]
+  end
+
```
ddottyler, line 36: Could you have done .to_i higher up in the code as soon as the roll is passed in so that you don't have to repeat it?

ddottyler (commented):
Overall, really impressive! You're getting the right outcome in a neat and concise way that doesn't seem to require a lot of code. 

it might be good to add in a feature that lets you see your total score frame by frame but otherwise really good!

PR #53, opened 2021-02-08
Bowling Challenge (Ruby)

HolsDuckett:


lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    elsif @spare == true
+      score = (bowl_1 * 2) + bowl_2
+    elsif
+      score = bowl_1 + bowl_2
+    end
+    return score
+  end
+
+
+  def strike?(bowl_1)
```
Mornevanzyl, line 46: Could be nice to rewrite these as ternary operators

Mornevanzyl (commented):
Solid and functional. Nice README. It's quite 'procedural' though, so it might be worth considering taking the bowling engine out and having the class(es) more internally focused.

lib/game.rb
bullhornfixie: Could use a new class for frame and scoring which keeps the control panel cleaner.

bullhornfixie: The use of a boolean to switch the spare status on and off seems the best approach.

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    if bowl_1 == 10
+      @strike = true
+    else
+      @strike = false
+    end
+  end
+
+  def spare?(bowl_1, bowl_2)
```
bullhornfixie, line 54: Could use a single line boolean here to condense code. Something like this:

roll1 != 10 && roll1 + roll2 == 10 ? @spare = true : @spare = false

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    if bowl_1 + bowl_2 == 10
+      @spare = true
+    else
+      @spare = false
+    end
+  end
+
+
+  def play_game
```
bullhornfixie, line 63: Short methods used with max 4-5 lines of code which is a +

PR #54, opened 2021-02-08
TT bowling challenge

twigz826:


domain_model.md
```diff
@@ -0,0 +1,68 @@
...
+Play game|Runs Game.play
+See frame score|Reads Score.frame_score
+See total score|Reads Score.total_score
+Input score|Changes game.score
+
+
+|Class|Properties(instance variables)|Actions(methods)|
+|---------|------------|-----------|
+|Game||play,score|
+|Score|frame_score, total_score|
```
KarstenFinlay, line 68: Really nice plan layout, will take this approach next challenge, did it work out for you?

lib/bowling_game.rb
```diff
@@ -0,0 +1,58 @@
...
+      else
+        @result += frame_score(roll_index)
+        roll_index += 2
+      end
+      @frame += 1
+    end
+    result
+  end
+
+  private
```
KarstenFinlay, line 36: Really like the idea of putting these methods into private. 

Also means fewer tests.

spec/bowling_game_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(game.score).to eq(0)
+    end
+
+    it "can roll all 4s" do
+      BowlingGame::MAX_ROLLS.times { game.roll(4) }
+      expect(game.score).to eq(80)
+    end
+
+    it "can roll a spare" do
+      game.roll(5)
```
KarstenFinlay, line 28: Would a spare not work if you put this in instead?:
`game.roll(9)`
`game.roll(1)`

KarstenFinlay: Wait I think I understand, why there is the extra `game.roll(3)`, no worries, me not reading ahead.

feature_test.rb
```diff
@@ -0,0 +1,27 @@
+2.2.3 :001 > g = Game.play
```
KarstenFinlay, line 1: Might need to walk this through with me? 
Is this just a feature test, to test the app in irb?

KarstenFinlay (commented):
Looks really good.

I think potentially, splitting the responsibilities of the `BowlingGame`, at the moment it does both scoring and the game. Perhaps this can be split into two classes `BowlingGame` and `BowlingScore`.

Really impressive README, I will probably use that layout for mine.

PR #58, opened 2021-02-08
Bowling-Challenge-Ruby-Glykeria

GlykeriaStr:
- Need to add more tests
- Edge cases ( can't hit more than 10 pins)
- Sort out when it's a spare and a strike one after the other


lib/scoreboard.rb
```diff
@@ -0,0 +1,82 @@
...
+    while @frames < 10
+      first_roll
+      strike? ? @frames += 1 : second_roll
+    end
+  end
+
+  def frame_10
+    first_roll
+    if strike?
+      bonus_roll
```
bateso88, line 34: I think this means the 10th frame will have 3 rolls if the first is a strike? I think if the first is a strike and the second isn't, it should only have two rolls. (Not 100% sure though!)

lib/scoreboard.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def strike?
+    @frame_score[@frames][0] == 10
+  end
+
+  def adding_strike_points
+    @frame_score.each do |frame, rolls|
+      @frame_score[frame] = [10 + (@frame_score[frame+1]).sum ]if rolls[0] == 10 && frame != 10
+    end
```
bateso88, line 61: I think this misses the edge case of three strikes in a row though? I.e. frame_1: Strike, frame_2: Strike, frame_3: strike... Frame_1 score should be 30 - I think at the moment it will be 20.

lib/scoreboard.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def adding_spare_points
+    @frame_score.each do |frame, rolls|
+      if rolls.sum == 10 && frame != 10
+        @frame_score[frame] = [10 + @frame_score[frame+1][0]]
+      end
+    end
+  end
```
bateso88, line 70: nice!

lib/scoreboard.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def total
+    @frame_score.each { |frame, rolls| @score << rolls }
+    @score.flatten.sum
+  end
```
bateso88, line 75: What does flatten do??

spec/scoreboard_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'scoreboard'
+
+
+describe Scoreboard do
+  subject(:scoreboard) { described_class.new }
+
+  describe '#first_roll' do
+    it 'stores the first roll of each frame' do
+      allow(scoreboard).to receive(:gets).and_return('6')
```
bateso88, line 9: nice!

bateso88: I didn't know you could do that with gets - handy

spec/scoreboard_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  describe '#second_roll' do
+    it 'stores the second roll of each frame' do
+      allow(scoreboard).to receive(:gets).and_return('6')
+      scoreboard.first_roll
+      allow(scoreboard).to receive(:gets).and_return('3')
+      scoreboard.second_roll
+      expect(scoreboard.frame_score).to eq(1 => [6, 3])
```
bateso88, line 21: I like the way frame_score is structured. Nice and concise

bateso88 (commented):
Great work! I'm amazed you kept everything so tidy! Definitely will be stealing some ideas for refactoring. A couple of point on functionality but on the whole - tidy, neat, clean, easy-to-read :)

PR #59, opened 2021-02-08
Bowling-Challenge 

louiswicks:


lib/.rspec
```diff
@@ -0,0 +1 @@
+--require spec_helper
```
acavalla, line 1: How come you have two .rspec files?

lib/bowling.rb
```diff
@@ -0,0 +1,51 @@
...
+    @rolls = []
+  end
+
+  def roll(pins)
+    @rolls << pins
+  end
+
+  def score
+    result = 0
+    rollIndex = 0
```
acavalla, line 13: Really like this rollIndex method, very elegant

lib/bowling.rb
```diff
@@ -0,0 +1,51 @@
...
+    @rolls[rollIndex] + @rolls[rollIndex + 1]
+  end
+
+  def strike?(rollIndex)
+    @rolls[rollIndex] == 10
+  end
+
+  def strikeScore(rollIndex)
+    @rolls[rollIndex] + @rolls[rollIndex + 1] + @rolls[rollIndex + 2]
+  end
```
acavalla, line 49: Think ruby uses snake_case for method names but i love the simplicity and SRPness of this

spec/bowling_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'bowling'
+
+describe 'A bowling game' do
+
+  before do
+    @game = Bowling.new
```
acavalla, line 6: Instead of @s and before do you can use subject { described_class.new } at the beginning and then use subject to refer to the game

acavalla: sorry if this isnt clear! we can chat this afternoon anyway

spec/spec_helper.rb
```diff
@@ -0,0 +1,100 @@
...
+  config.expect_with :rspec do |expectations|
+    # This option will default to `true` in RSpec 4. It makes the `description`
+    # and `failure_message` of custom matchers include text for helper methods
+    # defined using `chain`, e.g.:
+    #     be_bigger_than(2).and_smaller_than(4).description
+    #     # => "be bigger than 2 and smaller than 4"
+    # ...rather than:
+    #     # => "be bigger than 2"
+    expectations.include_chain_clauses_in_custom_matcher_descriptions = true
+  end
```
acavalla, line 29: what does this do?

spec/bowling_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    expect(@game.score).to eq 24
+  end
+
+  it 'can roll a perfect game' do
+    12.times{@game.roll(10)}
+    expect(@game.score).to eq 300
+  end
+
+
+end
```
acavalla, line 45: Doesn't prevent wrong scores being entered (no limit of 10 per roll)

acavalla: Maybe add an error message when the game is over?  And/or the functionality to output score during the game (atm it outputs an error)

spec/bowling_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'bowling'
+
+describe 'A bowling game' do
+
+  before do
+    @game = Bowling.new
+  end
+
+  it 'can make a new game' do
```
acavalla, line 9: This is sort of implied with the before do statement now so this test can be deleted

spec/bowling_spec.rb
acavalla: And/or a message when the game is over

acavalla (commented):
Amazing Louis!! love your code, so simple and elegant. Looking forward to learning from you during pairing! You could also update the readme

lib/bowling.rb
acavalla: Also can replace magic numbers (10) with constants

PR #60, opened 2021-02-08
incomplete challenge

HamishArro:
Sorry pete.


lib/bowling_score.rb
```diff
@@ -0,0 +1,44 @@
...
+    @frames = Array.new
+    @in_turn = false
+    @bowls = Array.new
+    @total_score = 0
+  end
+
+  def roll(pins_downed)
+    check_turn
+    turn(pins_downed)
+  end
```
PetarSimonovic, line 14: I think this is what I should have done, rather than processing frames, which made calculating scores for consecutive strikes a massive headache

PR #67, opened 2021-03-07
AJ's bowling challenge in Ruby

AJ8GH:
- specifications complete
- game + frame logic working
- scoring + bonus logic working


lib/frame.rb
```diff
@@ -0,0 +1,59 @@
...
+    @final = false
+    @remaining_pins = TOTAL_PINS
+  end
+
+  def score
+    rolls.sum + bonus
+  end
+
+  def add_roll(pins)
+    raise PinError if pins > remaining_pins
```
ianmcnicholas, line 19: An error is not raised when the user inputs a negative number or anything that isn't an integer.

ianmcnicholas:
Generally a fantastic piece of work.  Very neat and organised, with short methods and a very thorough README.

- Really enjoyed seeing the features suite of tests showcase full games.
- I was able to 'break' the game by using negative numbers.  Perhaps incorporate a way of preventing anything but positive integers being inputted?  NB have seen that you have already implemented this - great work!
- Good number of detailed commits.  The README is great - perhaps could be reordered?  Think about what a user would want to read first.
- Would be good to see a few more comments in the pull-request (so a colleague understands what they have been sent for example.)
- Finally, for the sake of a Makers student / beginner reading the code, it would be handy to have a comment with each method detailing what it for.

Enjoyed reviewing this Adam! Until next time...

lib/frame.rb
AJ8GH: good spot!

lib/score_board.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def running_total
+      scores = []
+      i = 0
+      while i < frames.count
+        scores << frame_scores[0..i].sum
+        i += 1
+      end
+      scores
+    end
```
alicelieutier, line 21: This feels inefficient as you are always counting back again from the start. Maybe you could store the accumulated score and add only the last frame each time?

spec/features/whole_games_spec.rb
```diff
@@ -0,0 +1,44 @@
+describe 'playing a whole game' do
```
alicelieutier, line 1: These are really great feature tests 🎉

spec/models/exceptions_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+  it 'raises pin error with custom message' do
+    expect { raise described_class }.to raise_error described_class, subject.msg
+  end
+end
+
+describe GameOverError do
+  it 'raises game over error with custom message' do
+    expect { raise described_class }.to raise_error described_class, subject.msg
+  end
+end
```
alicelieutier, line 11: I don't think you really need these. Testing should be for behaviour, here your error class is just storage no logic, so I probably wouldn't have tests for it. (controversial, maybe)

lib/frame.rb
```diff
@@ -0,0 +1,60 @@
+# knows if it's a strike, spare or over
+class Frame
+  TOTAL_PINS = 10
+
+  attr_reader :remaining_pins, :bonus, :rolls
```
alicelieutier, line 5: Should these really be publicly readable? Are you doing this mostly / only for testing?
It's useful when writing feature test to think about what the rest of the program will need to know about objects of this class. 
Here I would say they mostly need to:

**INPUTS**
- Create it
- Pass it a roll

**OUTPUTS**
- Know if it's over
- Know the score
- Know things like is it a strike or spare (maybe)
- Getting errors

That's all. So that's all the behaviour you need to expose and test. No testing the internal state. This way, it allows you more flexibility to change how that internal state works later on, while keeping the public interface and functionality.

spec/models/frame_spec.rb
```diff
@@ -0,0 +1,191 @@
+describe Frame do
```
alicelieutier, line 1: See my comments in the Frame class.

spec/models/game_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+  describe '#score' do
+    it 'calls on score_board' do
+      expect(score_board).to receive(:total_score).once
+      subject.score
+    end
+  end
+
+  describe '#current_frame' do
+    it 'holds an instance of frame' do
+      expect(subject.current_frame).to be frame
```
alicelieutier, line 17: This test is testing state not behaviour. Your test should not fix your implementation rigidly, and so they should only be testing inputs and outputs of your class.
outputs include how you class calls other classes and what they call them with (which you do). But the internal state should stay hidden.

spec/models/game_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    end
+
+    it 'becomes a new frame after frame is over' do
+      allow(frame).to receive(:over?).and_return(true)
+      expect(frame_class).to receive(:new).twice
+      subject.add_roll(3)
+    end
```
alicelieutier, line 24: I feel like this is still testing state/implementation. The reason is that how many times you call new on state is really down to your implementation. You could choose to do it after a roll has happened, or only before you add a roll. I don't think the tests should care.

spec/models/game_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+      10.times { subject.frames << frame }
+      expect { subject.add_roll(8) }.to raise_error GameOverError
+    end
+
+    context 'when frame is not over' do
+      it 'does not add current frame to frames' do
+        subject.add_roll(3)
+        expect(subject.frames).to be_empty
+      end
+    end
```
alicelieutier, line 43: Implementation again.

spec/models/game_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+  describe '#add_roll' do
+    it 'adds roll to current frame' do
+      expect(frame).to receive(:add_roll).with(1)
+      subject.add_roll(1)
+    end
+
+    it 'raises error when game is over' do
+      10.times { subject.frames << frame }
+      expect { subject.add_roll(8) }.to raise_error GameOverError
+    end
```
alicelieutier, line 36: These are good.

spec/models/game_spec.rb
alicelieutier: Not reading this full test file, as I feel I've given you a few indications of how to know whether a test is implementation/state or behaviour.

lib/frame.rb
alicelieutier: Reading more code, actually maybe rolls is needed as you show the rolls on your scorecard later.

alicelieutier (commented):
Overall:

Great README!
Good organisation of code.
Your code is easily readable.
Some minor naming comments.

Your tests are mostly good, but sometimes you test too much, by adding implementation/state tests instead of testing behaviour only.

See more comment below.

lib/frame.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  attr_writer :remaining_pins, :bonus
+
+  def final?
+    @final
+  end
+
+  def replace_pins
+    self.remaining_pins = TOTAL_PINS
+  end
```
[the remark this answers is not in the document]
AJ8GH: so to avoid this, I could use a different variable altogether for the case of the final frame, like you metioned above?

lib/frame.rb
```diff
@@ -0,0 +1,60 @@
...
+  def score
+    rolls.sum + bonus
+  end
+
+  def add_roll(pins)
+    pins = pins.abs
+    raise PinError if pins > remaining_pins
+
+    self.remaining_pins -= pins
+    replace_pins if final? && !over?
```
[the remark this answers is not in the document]
AJ8GH: Is this because instance variables should not be used for multiple purposes like this? or to avoid resetting the variable as you mention below?

I'm not sure I understand exactly, but would something like this be better?
```
  def initialize
    @bonus = 0
    @rolls = []
    @final = false
    @remaining_pins = TOTAL_PINS + ghost_frame
    @ghost_frame = TOTAL_PINS if final?
  end

File: spec/models/game_spec.rb (hunk 7 above)
Comment (AJ8GH) [reply spec/models/game_spec.rb:17]: cool, that makes sense. similar to how there's no need to test private methods - we only need to test for desired behaviour of the user?

In this case would you say we could remove the test for this method altogether?

File: lib/frame.rb [hunk 13]
```diff
@@ -0,0 +1,60 @@
...
+
+  def strike?
+    rolls.first == TOTAL_PINS
+  end
+
+  def add_bonus(pins)
+    self.bonus += pins
+  end
+
+  def final
```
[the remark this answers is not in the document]
AJ8GH: sure - make_final would be clearer

setting it at initialization time is also a nice idea, thank you

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+  attr_reader :frame_class
+  attr_writer :current_frame
+
+  def finish_frame
+    frames << current_frame
+    return if game_over?
+
+    self.current_frame = frame_class.new
+    current_frame.final if frames.count == 9
+  end
```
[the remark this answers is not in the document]
AJ8GH: Not sure really, just the way I designed it - I chose to keep it in current frame so that I could then add rolls to it there, etc. before pushing it. I suppose I could push it beforehand instead and do the same with `frames.last` instead

Do you think that would be better from a design point of view?

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+    frames[-2].add_bonus(pins) if second_strike_bonus?
+  end
+
+  def last_frame_bonus?
+    (frames.last.spare? || frames.last.strike?) && rolls_since <= 3
+  end
+
+  def second_strike_bonus?
+    frames[-2].strike? && rolls_since <= 2
+  end
```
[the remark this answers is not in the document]
AJ8GH: I like this Idea :) that could make the add bonus method a lot more simple

lib/score_board.rb
AJ8GH: I see what you mean. spent about 10 minutes last night on this and I did manage to refactor the method to be shorter:
```
def running_total
  frame_scores.map.with_index { |frame, i| frame_scores[0..i].sum }
end
```

although it still accumulates in the same way, so I think it's still inefficient. I'll try again if I find the time,

spec/models/exceptions_spec.rb
AJ8GH: got it. I removed them and my coverage didn't change, so I agree with you. thanks

lib/frame.rb
AJ8GH: yeah I see what you mean, I guess they don't need to be readable. I was kind of thinking ahead to the UI, that maybe this is all info I'd want to display to the user in game. However that can also be achieved through more purpose built public methods. thanks!

lib/frame.rb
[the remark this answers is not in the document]
alicelieutier: Updating them is absolutely fine. It's "resetting" that feels wrong. To me, it feels like what you initially use "remaining pins" for is one thing, and then you rest + use it for a different purpose, slightly. It's not like the frame is restarting.

spec/models/game_spec.rb
alicelieutier: Yes, absolutely. Also, maybe this is not a public method?

PR #69, opened 2021-03-08
Iain's bowling challenge

iainaitken:
See the README for notes and a description of the program


lib/frame.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class Frame
```
penguat, line 3: This class doesn't do very much, and I'd be tempted to merge it with the FrameScore class. This is a scorecard, so I feel the only model we really need for a frame is a model around scoring.

penguat (commented):
Hi Iain, Alex A here :)

I absolutely love your score display, and I think you've done really good work here.

As I believe you identified yourself, this is a little over-engineered - you didn't need all of these classes. That said, in different circumstances all of them could have been entirely appropriate. I think this additional weight is also really evident in the game class - with fewer classes I think you would also see opportunities in the game class to refactor and simplify it.

You also missed some opportunities in your display class to make things simpler and more readable. That's not major - it's well encapsulated and isolated, but I still think it's worth mentioning.

All in all, this is really good work. You identified the broad issues yourself and that's an important part of your ability to resolve them yourself and write better code in the long run.

PR #71, opened 2021-03-08
i don't roll on shabbas

jshorns:


lib/scorecard.rb
```diff
@@ -0,0 +1,65 @@
...
+    @final_score = 0
+  end
+
+  def add_frame(frame)
+    fail 'this game is over' if game_finished?
+    bonuses(frame)
+    @frames << frame
+    get_total if game_finished?
+    running_total
+  end
```
kerisic, line 17: Like the use of class extraction here so that frame is a separate class object.

lib/frame.rb
```diff
@@ -0,0 +1,22 @@
...
+    @final_score = @rolls.sum
+  end
+
+  def is_strike?
+    @rolls.first == 10
+  end
+
+  def is_spare?
+    !is_strike? && @rolls.sum == 10
+  end
```
kerisic, line 16: Awesome SRP here!

spec/scorecard_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'scorecard'
+require 'frame'
+
+describe Scorecard do
+  let(:frame) { double(:frame, :is_strike? => false, :is_spare? => false, :base_score => 7, :final_score => 7, :rolls => [4, 3])}
+  let(:frame2) { double(:frame, :is_strike? => false, :is_spare? => false, :base_score => 9, :final_score => 9, :rolls => [4, 5])}
+  let(:strike) { double(:frame, :is_strike? => true, :is_spare? => false, :base_score => 10, :final_score => 10)}
+  let(:spare) { double(:frame, :is_strike? => false, :is_spare? => true, :base_score => 10, :final_score => 10 )}
```
kerisic, line 8: Good use of doubles and method stubbing here

kerisic (commented):
Nice encapsulation done and clean, easy to understand the scoring rules/logic. Wonder if it can also include some edge cases around inappropriate number of arguments, or sum for frame. Great job as a whole though!

PR #75, opened 2021-03-08
lets go bowling

jgumoes:


lib/score.rb
```diff
@@ -0,0 +1,113 @@
+require_relative 'badscore_error'
+
+class Score
+  attr_reader :name, :scores
+
+  def initialize name = 'Sisyphus'
+    @name = name   
+    init_scores()
+  end
+
```
iainaitken, line 10: I'm wondering whether a couple more classes might help improved readability - we could make it more intuitive than score.name giving the player's name. A user class could help here?

README.md
```diff
@@ -1,65 +1,80 @@
...
-### Perfect Game
-
-A Perfect Game is when the player rolls 12 strikes (10 regular strikes and 2 strikes for the bonus in the 10th frame). The Perfect Game scores 300 points.
-
-In the image below you can find some score examples.
-
-More about ten pin bowling here: http://en.wikipedia.org/wiki/Ten-pin_bowling
-
-![Ten Pin Score Example](images/example_ten_pin_scoring.png)
+# Quickstart
```
iainaitken, line 1: Generally, I think this program is really well implemented, and it works, which is the main thing! The only thing I think could be improved is readability. It doesn't seem very ruby-esque - it reminds me more of programs written in a lower-level language like C, and it's not very easy for someone not familiar with the code to read. The comments and README help, though.

PR #76, opened 2021-03-08
Started project. Player, Game & Frame objects created. Associated met…

kasey-purvor:
…hods added. Still to do: create a game loop for the command line.


charlierdm:
Coder review:

- no commits during the project? 
- no readme, would be nice to see how to use the programme.
- clear class responsibilities.
- doesn't need a 2nd player, just a scorecard for 1 player.
- good testing on the player class.
- can see where the project is going as it is clearly written.

PR #78, opened 2021-04-03
complete bowling challenge in ruby

loushark:
Still left To-Do:
[ ] roll 2 cannot be played if roll 1 is a strike
[ ] value of roll 2 cannot be a number where, when added to roll 1 value, it will equal more than 10
[ ] logic of more than one strike in a row

See README for more info and IRB walkthrough


README.md
```diff
@@ -58,8 +35,91 @@ A Gutter Game is when the player never hits a pin (20 zero scores).
...
 A Perfect Game is when the player rolls 12 strikes (10 regular strikes and 2 strikes for the bonus in the 10th frame). The Perfect Game scores 300 points.
 
-In the image below you can find some score examples.
-
-More about ten pin bowling here: http://en.wikipedia.org/wiki/Ten-pin_bowling
-
-![Ten Pin Score Example](images/example_ten_pin_scoring.png)
+---
+
+### Modelling
```
fg24davies, line 40: This is a great addition to a README! I have learnt a lot from this so thanks!

README.md
```diff
@@ -58,8 +35,91 @@ A Gutter Game is when the player never hits a pin (20 zero scores).
...
+|                |                                | end_of_game                                    |
+|  Scorecard     |   @scorecard = []              | update_bonus(players_score)                    |
+|                |   @roll_1_score = 0            | update_scorecard(current_frame, strike, spare) |
+|                |   @roll_2_score = 0            | running_total(current_frame)                   |
+|                |   @roll_3_score = nil          |                                                |
+|                |   @bonus_score = 0             |                                                |
+|                |   @total_game_score = 0        |                                                |
+
+
+### irb
```
fg24davies, line 64: Really nice layout of how your program works in IRB. I find it's really useful to have this before looking at the ruby file, for better understanding!

spec/bowling_game_spec.rb
```diff
@@ -0,0 +1,126 @@
+require 'bowling_game'
```
fg24davies, line 1: your tests were really clear and thorough. I like the context blocks. Your use of doubles has also made it really clear to follow!

lib/bowling_game.rb
```diff
@@ -0,0 +1,112 @@
+require_relative 'scorecard'
```
fg24davies, line 1: You were right that it was a little bit hard to make any meaningful comments without the challenge myself. 
I am able to follow most of code and what your methods do. I think you have followed SRP well, the two classes splitting responsibility of rolling and scoring is nice! Most of your methods are really short as well. I like how you have split it in to roll 1 and roll 2 for clarity also!

fg24davies (commented):
Hi Lou, you were right in that it was hard to comment anything meaningful without further context. 
I thought your testing was really really good! As was how you created the interaction with game from IRB, that makes a lot of sense. I think printing to the console after an important event also helps you! 
There is so much logic in this and so I thought you did really well to keep your methods short where you could.
Really nice job well done :)

lib/bowling_game.rb
```diff
@@ -0,0 +1,112 @@
...
+      update_roll_1_score(players_score)
+      update_roll_2_score(0)
+      update_scorecard
+      @strike = true
+      "Strike!"
+    else
+      update_roll_1_score(players_score)
+      "Nice roll! Let's Roll again!"
+    end
+  end
```
StuBehan, line 35: I feel like this is a bit unwieldy, I think there are elements that could be extracted into helper methods easily, especially where there are reused.

```suggestion
  def roll_1(players_score)
    update_bonus(players_score)  # extract all the logic from here to a helper method

    if players_score < 10 && @current_frame == 10 # extract the current_ftame == 10 to some method like tenth_frame?
      update_roll_1_score(players_score)
      update_roll_2_score(0)
      update_roll_3_score(0)
      update_scorecard
    elsif players_score == 10 && @current_frame < 10 # then use !tenth_frame?
      update_roll_1_score(players_score)
      update_roll_2_score(0)
      update_scorecard
      @strike = true
      "Strike!"
    else
      update_roll_1_score(players_score)
      "Nice roll! Let's Roll again!"
    end
  end
```

lib/bowling_game.rb
```diff
@@ -0,0 +1,112 @@
...
+    @scorecard.roll_1_score = players_score
+  end
+
+  def update_roll_2_score(players_score)
+    @scorecard.roll_2_score = 0
+  end
+
+  def update_roll_3_score(players_score)
+    @scorecard.roll_3_score = 0
+  end
```
StuBehan, line 94: do we need to pass an argument if we arent using it?

lib/bowling_game.rb
```diff
@@ -0,0 +1,112 @@
...
+      @spare = true
+      return "Spare!"
+    end
+    "Great Job! That's the end of this frame"
+  end
+
+  def roll_3(players_score)
+    if @current_frame < 10
+      return "You can only roll a third time in the 10th frame"
+    end
```
StuBehan, line 64: do we ever use this? does the logic bring us to a 3rd roll outside of the tenth frame?

StuBehan (commented):
I think you've done a good job of this challenge, there isn't too much to reflect on, just maybe extracting some stuff into helper methods, have you ran Rubocop on it? Maybe take a look at PReek to see what that comes out with regards to your code and its 'smells' (`gem install preek`, then just run `preek ./lib`)

lib/bowling_game.rb
loushark: I guess this was more for irb purposes where if a player tried to roll a third time they couldn't, and would be notified why - maybe this could be more of an error return? But when applied to a web app, this probably wouldn't be needed

PR #80, opened 2021-04-04
bowling

Emmapr123:


README.md
```diff
@@ -33,33 +9,78 @@ In order to do this, you may pay particular attention to the following:
...
-    1, 9, 10 in the 10th frame gives 20 points (10 points for the regular spare and 10 points for the bonus).
-
-### Gutter Game
-
-A Gutter Game is when the player never hits a pin (20 zero scores).
-
-### Perfect Game
+## Bowling Challenge
+
+#### User stories
```
inbarvernia, line 14: I like the inclusion of user stories here - makes it really clear what the code is going to do, even for people who might not know how bowling works.

lib/frame.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class Frame
```
inbarvernia, line 3: Great idea to have frames as its own class! Very clear and concise methods, good use of encapsulation and SRP.

spec/frame_spec.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'frame'
+
+describe Frame do
+  let(:frame) { Frame.new }
+  describe '#rolls' do
+    it 'should go to the next roll' do
```
inbarvernia, line 8: I'm finding this it statement a little bit unclear: it looks like the test is testing whether a new frame starts out with no rolls in it, but the statement implies it should be testing if it goes to the next roll; maybe either the test could be changed or the it statement could be reworded?

spec/frame_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      expect(frame.is_spare?).to eq true
+    end
+    it 'should know a spare was not rolled' do
+      frame.roll(4)
+      frame.roll(3)
+
+      expect(frame.is_spare?).to eq false
+    end
+  end
```
inbarvernia, line 48: Really good, clear tests.

lib/frame.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def is_strike?
+    @rolls.first == 10
+  end
+
+  def is_spare?
+    @rolls.length == 2 && count == 10
+  end
+
+  def count
```
inbarvernia, line 22: Is this method necessary, or would it be just as good to call `@rolls.sum` directly above? Unless it also needs to be called by another class below, in which case, that makes sense!

lib/score_card.rb
```diff
@@ -0,0 +1,82 @@
...
+    unless final_frame
+      @score = @frames.map.with_index do |frame, index|
+        bonus = 0
+        bonus += frame.rolls.first if previous_two_frames_are_strikes(index)
+        if previous_frame_is_strike(index)
+          bonus += frame.count
+        elsif previous_frame_is_spare(index)
+          bonus += frame.rolls.first
+        end
+        bonus + frame.count
```
inbarvernia, line 36: Ah, okay, now I understand why you needed to define it as a separate method! Disregard my earlier question. :)

spec/frame_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+describe Frame do
+  let(:frame) { Frame.new }
+  describe '#rolls' do
+    it 'should go to the next roll' do
+      expect(frame.rolls).to eq []
+    end
+  end
+
+  describe 'roll' do
```
inbarvernia, line 13: Left out a # in `#roll`

spec/score_card_spec.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+require 'score_card'
+
+describe ScoreCard do
+  let(:bowl) { ScoreCard.new }
+  let(:frame) { double :frame }
+  describe '#frames' do
+    it 'should be 1 by default' do
```
inbarvernia, line 9: I think this it statement could be clearer - maybe something like 'should exist', 'should be an empty array' / 'should be empty', 'should be empty before any frames have been completed'?

spec/score_card_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+      expect(bowl.frames).to eq []
+    end
+  end
```
inbarvernia, line 12: Perhaps also include a test to check that frames are added to it after the end of a frame? Or is this implicitly tested by the tests for `#roll` below?

lib/score_card.rb
```diff
@@ -0,0 +1,82 @@
...
+
+class ScoreCard
+  attr_reader :frames
+
+  def initialize
+    @frames = []
+    @score = 0
+  end
+
+  def play(user_input)
```
inbarvernia, line 11: Maybe `user_input` here and below can be renamed to something more descriptive, like number_of_pins, pins_knocked, etc.?

lib/score_card.rb
```diff
@@ -0,0 +1,82 @@
...
+    final_frame ? tenth_round(user_input) : roll(user_input)
```
inbarvernia, line 12: I like this ternary operation for determining what type of turn this is - clear and efficient!

spec/frame_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+    it 'should know when it is not a strike' do
+      frame.roll(4)
+
+      expect(frame.is_strike?).to eq false
+    end
+  end
+
+  describe '#is_spare?' do
+    it 'should know a spare wass rolled' do
```
inbarvernia, line 36: Typo - 'wass'

lib/score_card.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def roll(user_input)
```
inbarvernia, line 15: If you want to be able to test this independently from the Frame class using a double, you could use dependency injection by adding an argument with default value, e.g. `frame = Frame.new`. This wouldn't affect the functionality of the code, but it would allow you 'feed' a double into the method in the tests. Don't think it's entirely necessary, but it's an option.

spec/score_card_spec.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+require 'score_card'
+
+describe ScoreCard do
+  let(:bowl) { ScoreCard.new }
+  let(:frame) { double :frame }
```
inbarvernia, line 7: At the moment, I think this line is unnecessary as you are not using a double in the tests - the methods interacting with frame objects are calling actual frame objects directly.

spec/score_card_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+      expect(bowl.frames.length).to eq 1
+    end
+    it 'saves each frame seperately' do
+      bowl.roll(10)
+      bowl.roll(6)
+      bowl.roll(1)
+      bowl.roll(7)
+
+      expect(bowl.frames.length).to eq 3
+    end
```
inbarvernia, line 28: Really good test - thorough.

inbarvernia (commented):
Really good! Good use of classes, concise methods with clear responsibilities, and good support methods. Code was very clear and easily understandable, and testing was thorough. :)

lib/frame.rb
Emmapr123: I was thinking about this as well, I guess this one isn't necessary, it might be more DRY, as I use it to calculate bonuses, where my function now does not need to do the rolls.sum

lib/score_card.rb
Emmapr123: Yes, good shout! I was thinking of doing something like that too, but didn't get around to it yet!

spec/frame_spec.rb
Emmapr123: Good call, I was writing the test with the intention to test it going to the next one, and then decided it would be better practice to see if it is initialized as an empty array first, but forgot to change the it statement.

PR #82, opened 2021-04-05
Partial submission

inbarvernia:
See README for details on features implemented and not implemented. I would particularly appreciate feedback on whether it seems like the tests are testing state instead of behaviour, if the method and variable naming is clear enough, and if the `roll` method is okay as it is, or if it's still doing too much. Many thanks! :)


README.md
```diff
@@ -14,52 +13,27 @@ Count and sum the scores of a bowling game for one player. For this challenge, y
...
-A Perfect Game is when the player rolls 12 strikes (10 regular strikes and 2 strikes for the bonus in the 10th frame). The Perfect Game scores 300 points.
-
-In the image below you can find some score examples.
-
-More about ten pin bowling here: http://en.wikipedia.org/wiki/Ten-pin_bowling
-
-![Ten Pin Score Example](images/example_ten_pin_scoring.png)
+## My implementation
+
+To fulfill this challenge, I created a BowlingScore class, each instance of which being the scorecard for one player. Within this class, I've defined an initialize method and attribute reader, and the method `#roll` which takes the number of pins knocked down as an argument and updates the score (which it also returns), relying on some private supporting methods. I have implemented the following features:
```
Emmapr123, line 18: Nice summary of the challenge, a great practice to explain it to yourself before you get started!

lib/bowling_score.rb
```diff
@@ -0,0 +1,55 @@
...
+    @total_score = 0
+    @frame_number = 1
+    @current_frame = []
+    @frames = {}
+  end
+
+  def roll(pins)
+    @current_frame << pins
+    @total_score += pins * number_of_bonuses
+    if @current_frame.length == 2 || pins == 10
```
Emmapr123, line 14: This is generally a good one, but it wont allow for the tenth frame to have a third roll, how would you approach that?

lib/bowling_score.rb
```diff
@@ -0,0 +1,55 @@
...
+  def last_frame
+    @frame_number - 1
+  end
+
+  def spare_bonus?
+    @frames[last_frame].sum == 10 && @frames[last_frame].length == 2 && @current_frame.length == 1
+  end
+
+  def strike_bonus?
+    @frames[last_frame].length == 1
```
Emmapr123, line 31: would you need a last frame function? wouldn't [index - 1] do the job?

lib/bowling_score.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def double_strike_bonus?
+    frame_before_last = last_frame - 1
+    @frames[frame_before_last].length == 1 && strike_bonus?
```
Emmapr123, line 36: and [index - 2] here?

lib/bowling_score.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def number_of_bonuses
+    if @frame_number > 2 && double_strike_bonus?
+      2
```
Emmapr123, line 41: cool way of dealing with bonusses!

spec/bowling_score_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+          player1.roll(10)
+          expect { player1.roll(10) }.to change { player1.total_score }.by(20)
+          # Increase of 20 is made up of 10 (overall frame score for second strike) and 10 (first bonus for first strike)
+        end
+
+        it 'adds two strike bonuses simultaneously for first roll after two strikes' do
+          player1.roll(10)
+          player1.roll(10)
+          expect { player1.roll(8) }.to change { player1.total_score }.by(16)
+          # Increase of 16 is made up of 8 (second bonus for first strike) and 8 (first bonus for second strike)
```
Emmapr123, line 90: would this still work with 3 or 4 in a row?

spec/bowling_score_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        end
+      end
+    end
+  end
+end
```
Emmapr123, line 95: Altogether, I like your approach! Looks like a good bit of code, curious to see how you would have done the entire challenge.

lib/bowling_score.rb
inbarvernia: Yes, I hadn't gotten quite that far! I thought I would probably define a separate predicate method (e.g. `tenth_frame?`) and add a conditional line of `if tenth_frame?` then execute something different (probably a separate method, or have a parent method calling either this or the tenth frame method as appropriate).

spec/bowling_score_spec.rb
inbarvernia: Yes, I manually tested it in IRB, but you're right that I should also include a unit test for that.

lib/bowling_score.rb
inbarvernia: Thank you. :)

PR #83, opened 2021-04-06
Bowling scorecard calculator

taran314:
Code can calculate the scores at the end of the game but did not have time to complete the method which updates scores ASAP. Focus was to get the scoring working and tidy up next weekend! Requires a lot of refactoring: another class - Frame, new methods to keep code DRY and remove extra variables. Any tips would be appreciated!


lib/bowling.rb
```diff
@@ -0,0 +1,128 @@
...
+    else
+      spares << frame if knocked_down == 10
+      @frame += 1
+      @knocked_down = 0
+    end
+    update_scores
+    change_roll
+  end
+
+  # Allows bowling to be played on irb - uncomment lines 127 & 128
```
louballantyne, line 32: Love the comments - they're super helpful!

README.md
```diff
@@ -1,10 +1,32 @@
...
+-----
+- Started with basic methods and implementing spares (w/o special 10th frame)
+- Then implemented logic for strikes and got scores for each frame correct
+- Implemented 10th game logic for normal games
+- Checked edge cases for perfect and gutter games
+- Created function which updates scores at end of game and play function
+- In process of implementing update_scores function which updates a frame's score ASAP, works for a perfect game but nothing else...
+- Code currently does not defend against wrong inputs - will break the calculator.
+- Code needs a lot of refactoring and new classes, any advice would be appreciated!
+
```
louballantyne, line 21: Great README

lib/bowling.rb
```diff
@@ -0,0 +1,128 @@
...
+      return strike if knocked_down == 10
+    else
+      spares << frame if knocked_down == 10
+      @frame += 1
+      @knocked_down = 0
+    end
+    update_scores
+    change_roll
+  end
+
```
louballantyne, line 31: I really like that you've extracted 'update_scores' into its own method, and the way the scores are displayed in the scorecard.

lib/bowling.rb
```diff
@@ -0,0 +1,128 @@
...
+    end
+    if @first_roll && current_frame > 2
+      if current_frame == 3
+        @final_scores[current_frame - 2] << @scores[current_frame - 2].sum        
+      else
+        @final_scores[current_frame - 2] << (@scores[current_frame - 2] + @final_scores[current_frame - 3]).sum
+      end
+    end
+  end
+
```
louballantyne, line 73: To reduce the complexity of 'if' statements and improve readability, you could extract parts of these methods into further methods.

spec/bowling_spec.rb
```diff
@@ -0,0 +1,177 @@
+require 'bowling'
+
```
louballantyne, line 2: Your testing is generally really thorough and clear as to what you're testing for.

lib/bowling.rb
```diff
@@ -0,0 +1,128 @@
+class Bowling
```
louballantyne, line 1: I think I'd consider the responsibilities of this class: At the moment, it's responsible for co-ordinating the game (calling methods at the right time etc), interacting with the user to obtain rolls and outputting the score, creating the score sheet and determining what should happen in each frame based on user input. As you said, you could definitely extract this into one more class, but it might be worth thinking about two further classes.

lib/bowling.rb
```diff
@@ -0,0 +1,128 @@
...
+  def roll(pins)
+    @knocked_down += pins
+    return tenth_frame(pins) if frame == 10
+    scores[frame] << pins
+    update_bonus_points(pins)
+    
+    if first_roll 
+      return strike if knocked_down == 10
+    else
+      spares << frame if knocked_down == 10
```
louballantyne, line 24: Rather than keeping a record of strikes and spares, then determining additional points based on what strikes and spares there are, perhaps you could just keep track, for each frame, of how many additional rolls should be factored in?

louballantyne (commented):
I love the way your code runs. It's really clear for the user what they've scored and why, and how bonus points are factored in. As you suggested, I think you need to refactor into at least one additional class, and to refactor your if statements into further methods to make your code more readable.

PR #84, opened 2021-04-06
working

ralphbartley:
So initially I wrote this challenge up, committing as I went with TDD, but never once pushed to this github repo.  Then I lost my directory with the project folder inside due to a mounting error in my virtual setup (lesson learned -- PUSH!).  So what I've got here is what I managed to write back down in about 2 hours.  As a result I didn't re-write any of the tests, just double checked functionality by quickly testing the behaviour manually.  This also means I lost my nice score printout formatting, but the basics are here.

My main question: what's the best way to handle the error handling seen in my /lib/scorecard.rb last_frames_specials ? Fundamentally I'm checking whether there is an object present so that I don't get an error on calling methods on nil when the arrays are empty.  Just reeks of code smell, so I'm curious there.

Spare and Strike classes fundamentally the same with 1 thing changed -- I chose not to make them related as classes, should I have?


ralphbartley:
updated the repo with a change that fixes the last frames that I had missed from before -- didn't do a new pull request

PR #86, opened 2021-04-06
Bowling

TataSher:
Created 2 classes: Frame and Game.
Frame is supposed to generate frames (and create a variable 'frame' that is a hash which contains rolls).
Game is supposed to generate games (and create a variable 'frames' that is an array and has all the 'frame' hashes of the game).
Game was also supposed to generate score by analyzing the 'frames' variable and assigning different scores to 'strike' and 'spare'. I still need to write this one.
I got stuck at writing a test for the frame double in the game_spec.rb :/


lib/frame.rb
```diff
@@ -0,0 +1,44 @@
...
+    @frame = {}
+  end
+
+  def first_roll(knocked_pins1)
+    @knocked_pins = knocked_pins1
+    if knocked_pins == 10
+       @frame = {frame_id: @frame_id += 1, first_roll: 'strike' }
+    else
+      @frame = {frame_id: @frame_id += 1, first_roll: knocked_pins1 }
+    end
```
jackmcc08, line 16: What is @frame_id doing? 
As this is common to both strings, you could pull it out of the conditional - avoids duplication.

spec/game_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    describe '#turn' do
+      it 'returns the received frame' do
+        game = Game.new
+        frame_double = double
+        allow(frame_double).to receive(:frame).and_return( { frame_id: 2, first_roll: 6, second_roll: 2 } )
+        game.turn(frame_double)
+        expect(game.frames).to eq[{ frame_id: 2, first_roll: 6, second_roll: 2 }]
+      end
+    end
```
jackmcc08, line 19: I see what you were trying to do here - I think you just need to initialize the double differently? I have not seen it done this way.

spec/frame_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      expect(frame.frame).to eq({ frame_id: 1, first_roll: 'strike' })
+    end
+
+    it "records spares" do
+      frame.first_roll(1)
+      frame.second_roll(9)
+      expect(frame.frame).to eq({ frame_id: 1, first_roll: 1, second_roll: 'spare' })
+    end
+  end
+end
```
jackmcc08, line 51: tests are clear and they contain the test set up in the test which makes them easier to understand what they are testing.

lib/frame.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def over
+    if @knocked_pins > 10 
+       raise 'Wrong number - re-type!'
+    end
+  end
+  
+end
+
+
```
jackmcc08, line 44: code is well named, makes sense what you were trying to achieve here. 

Couple of thoughts on this: 
- you could simplify some of the conditionals to be one liners
- your error catcher for if a wrong number is entered could be at the top of the first and second roll methods so that no logic is executed before an error is caught.

lib/game.rb
```diff
@@ -0,0 +1,15 @@
...
+  attr_reader :frames
+
+  def initialize
+    @frames = []
+  end
+
+  def turn(frame)
+    @frames << frame.frame
+  end
+
```
jackmcc08, line 14: What was frame.frame trying to do on line 12?

jackmcc08 (commented):
Overall code is well named and simple to understand. 

I wrote a few thoughts about how you could adjust your code.

PR #98, opened 2021-05-04
abdur 

Abd27:


lib/scorer.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+class Score
```
kerrimcm, line 3: Did you write tests for this class?

lib/scorer.rb
```diff
@@ -0,0 +1,48 @@
...
+  attr_reader :score_card, :total_score
+
+  def initialize
+    @total_score = 0
+    @score_card = []
+  end
+
+  def scorer(score1, score2)
```
kerrimcm, line 11: Might be good to extract these into their own methods which DRY up the code

lib/scorer.rb
```diff
@@ -0,0 +1,48 @@
...
+    else
+      set[:roll1] = score1
+      set[:roll2] = score2
+      @total_score += (score1 + score2)
+      set[:total_score] += (score1 + score2)
+      @score_card << set
+    end
+  end
+
+  def update_bonus
```
kerrimcm, line 35: Curious to know what this method does in the context of your program, I have an idea but I'm not clear on it looking at the code

lib/scorer.rb
```diff
@@ -0,0 +1,48 @@
...
+    (0...@score_card.length).each do |i|
+      @score_card[i - 1][:total_score] += @score_card[i][:roll1] if @score_card[i - 1][:spare?] == true
+      @score_card[i - 1][:total_score] += @score_card[i][:roll1] if @score_card[i - 1][:strike?] == true
+      @score_card[i - 1][:total_score] += @score_card[i][:roll2] if @score_card[i - 1][:strike?] == true
+    end
+  end
+
+  def display
```
kerrimcm, line 43: Could be good to put something like this in its own `Game` class of sorts to follow SRP more closely, since this class should just be responsible for the scoring of the game.

PR #99, opened 2021-05-04
Challenge progress

Mintar89:


noarfarber (commented):
Well-written code overall. Your code is clear and easy to understand, so well done!
The main thing I would suggest taking into consideration is that the 'score' function does not work in IRB, so maybe you find it worths fixing.

PR #103, opened 2021-05-04
Muhammad Ahmad - Bowling in ruby

ahmad047:


lib/scorecard.rb
```diff
@@ -0,0 +1,54 @@
...
+      if @scorecard[@frame_number - 1][:strike?] == true
+        @scorecard[@frame_number - 1][:total_score] += score
+      end  
+      @scorecard[@frame_number][:spare?] = true if @scorecard[@frame_number][:roll_1_score] + score == 10
+      @frame[:roll_2_score] += score
+      @frame[:total_score] += score
+      @scorecard[@frame_number] = @frame
+      update_frame
+    end
+    @scorecard.drop(1) if @frame_number == 1
```
jasonrowsell, line 41: Break this method down into separate methods. It'll make it easier to debug and alter functionality. 

- A method to calculate the score at the end of the game can be extracted from this, it can iterate over an array of rolls and add to an initialized result var. Within this method, include private methods that check if index & index+1 == 10... this will be a spare, use this to add a bonus to the result var. A private method to check if index == 10... this will be a strike, do the same.

- With this extraction, a method can be set up just to add rolls(a num from 1-10) into an array each time its called.

- TL:DR:  consider separating methods for easier logic and functionality :)

lib/scorecard.rb
```diff
@@ -0,0 +1,54 @@
+class Scorecard
```
jasonrowsell, line 1: Should a scorecard class be responsible for adding a score, storing frame numbers, rolling?

lib/scorecard.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def initialize
+    @roll = 1
+    @scorecard = [{}]
+    @frame_number = 1
+  end
+
+  def add_score(score)
+    if @roll == 1
+      if @scorecard[@frame_number - 1][:strike?] == true || @scorecard[@frame_number - 1][:spare?] == true
```
jasonrowsell, line 12: not 100% sure but when using || you don't need to state that they == true. the conditions checks for a 'true' evaluation

PR #105, opened 2021-05-29
week 5 end of unit challenge submission

rdupplaw:


lib/frame.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def strike?
+    @rolls.first == FRAME_PINS
+  end
+
+  def spare?
+    return false if @rolls.length < 2
+
+    !strike? && @rolls[0] + @rolls[1] == FRAME_PINS
```
GunelMC, line 35: !strike? && @rolls.sum == FRAME_PINS && @rolls.length == 2

PR #120, opened 2021-07-05
Chuka's Bowling Challenge

coo990:
Current score and number or rolls instance variables added, needs working on total score and frame counting. Currently all RSpec tests pass, need to add feature test as well in the future.


spec/bowling_score_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  context '#gutter' do
+    it 'gutters the ball' do
+      bowling.gutter
+      expect(bowling.current_score).to eq(0)
+    end
+  end
+  
+  context '#spare' do
+    it 'hit a spare' do
```
mikey-roberts, line 29: Good testing in place to check if result is a spare

lib/bowling.rb
```diff
@@ -0,0 +1,30 @@
+class BowlingScore
+  attr_reader :current_score
+
+  def initialize
+    @number_of_rolls = 0
```
mikey-roberts, line 5: like the use of incremental number of rolls to keep track of how many shots have been taken

lib/bowling.rb
```diff
@@ -0,0 +1,30 @@
...
+  def hit(number)
+    @number_of_rolls += 1
+    @current_score += number
+      if @current_score == 10
+        spare
+      end
+  end
+
+  def gutter
+    @current_score += 0
```
mikey-roberts, line 18: Nice break down of individual states of the game with separate methods for gutter, spare, etc. 

Missing logic but initial methods look to be on the right track 👍🏼

spec/bowling_score_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it 'current score to equal 9' do
+      bowling.hit(5)
+      bowling.hit(4)
+      expect(bowling.current_score).to eq(9)
+    end
+  end
+
+  context '#gutter' do
+    it 'gutters the ball' do
+      bowling.gutter
```
mikey-roberts, line 23: Current score is a good way to track the hit of the throw

PR #131, opened 2021-08-30
Weekend Challenge - Week 5

ClaimingCookie5:
I'm not happy with how I've written this.

I would like feedback on my #calculate_score method and clarity on whether the tests I have written are rigorous.

Thanks.


lib/scoring_bowling.rb
```diff
@@ -0,0 +1,67 @@
...
+  def initialize
+    @rounds = []
+    @frame = 0
+    @score = 0
+  end
+
+  def add_bowl(*bowls)
+    raise 'Game Over, no more scores can be added' if @rounds.length == MAX_FRAMES
+    
+    @rounds << { frame: frame_count, bowls: [*bowls] }
```
yousufmohamed17, line 15: It would be good if there was a way to add scores for individual rolls rather than entire frames.

lib/scoring_bowling.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  private
+
+  def calculate_all_scenarios(which_frame, arr)
+    if strike?(round(which_frame))
+      calculate_strike(which_frame, arr)
+    elsif spare?(round(which_frame))
+      calculate_spare(which_frame, arr)
+    else
+      arr << round(which_frame)
```
yousufmohamed17, line 37: Can we access this array from here?

spec/scoring_bowling_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  describe '#add_bowl' do
+    it 'expected to add the bowl to rounds array' do
+      subject.add_bowl(9, 0)
+      expect(subject.rounds).to include(frame: 1, bowls: [9, 0])
+    end
+
+    it 'expected to be able to take two bowls' do
+      subject.add_bowl(9, 1)
+      expect(subject.rounds).to include(frame: 1, bowls: [9, 1])
+    end
```
yousufmohamed17, line 20: I think these two tests could be combined.

lib/scoring_bowling.rb
```diff
@@ -0,0 +1,67 @@
...
+  def calculate_score
+    current_frame = 0
+    arr = []
+    # Needs to be refactored, unsure how.
+    until current_frame == rounds.length
+      calculate_all_scenarios(current_frame, arr)
+      current_frame += 1
+    end
+    @score += arr.flatten.sum
+  end
```
yousufmohamed17, line 27: Is it possible to get a running score throughout the game with this method?

yousufmohamed17 (commented):
Overall I think it would be good to add a score class or something similar because your game class is currently responsible for a lot.

lib/scoring_bowling.rb
ClaimingCookie5: poorley named parameter. I should make it clear what it is.

PR #134, opened 2021-08-30
Pin Pals - Lizzy B Bowling Challenge Ruby

eballantine:
I had the choice today of starting afresh but decided to keep my false starts in my commit history - as a lesson learnt. 

I think my concept of rolls remaining on a frame before it's scored is clean logically, but detached from the real world understanding of regular frame vs bonus points. If we wanted to show the running game state on a screen, you would also need a concept of frame results following each roll.


lib/frame.rb
```diff
@@ -0,0 +1,35 @@
+class Frame
+  attr_reader :frame_score, :current_roll, :rolls_remaining
+  attr_accessor :closed
+
+  def initialize
```
necsanuradu, line 5: Everything looks good the methods are short and concise(names are telling the story, what they do)

PR #138, opened 2021-08-31
Fiona's Weekend Challenge Pull Request

oddpotato:
Plus side, logic was written for a full game. Downside, testing just got lost along the way


scorecard_vanilla.rb
```diff
@@ -0,0 +1,55 @@
+# This is the 'vanilla' scorecard. It doesn't do anything fancy, 
+# it has no classes, it just does what is asked for on the assignment
+
+def game
+    p 'welcome to bowling!'
```
edpackard, line 5: I like the way you've implemented a clear, straightforward game loop here

scorecard_vanilla.rb
```diff
@@ -0,0 +1,55 @@
...
+        score(10,0)
+    else 
+        p 'please enter your second score'
+        second_bowl = gets.chomp().to_i
+        score(first_bowl, second_bowl)
+    end
+end
+
+def score(x,y)
+    running_score = []
```
edpackard, line 31: does this score method calculate the score for the frame, or the total score for the entire game?

lib/scorecard.rb
```diff
@@ -0,0 +1,23 @@
...
+require './lib/last_round'
+
+class Scorecard
+
+    attr_reader :score_card, :total_score
+
+    def game
+        @round_count = 0
+        @score = Score.new
+        while @round_count < 9
```
edpackard, line 12: good translation of your 'vanilla' game logic into this class

lib/last_round.rb
```diff
@@ -0,0 +1,27 @@
...
+            p 'STRIKE!!!'
+        end
+        p 'What was your second score?'
+        @round2 = gets.chomp().to_i
+        if @round2 == 10
+            p 'STRIKE!!!'
+        elsif @round1 + @round2 == 10
+            p 'SPARE!!!'
+        end
+        if @round2 == 10 || @round1 + @round2 = 10
```
edpackard, line 20: good use of OR logic

lib/score.rb
```diff
@@ -0,0 +1,66 @@
...
+    attr_reader :first_bowl, :second_bowl, :total_score, :score_card
+
+    def initialize
+        @score_card = []
+        @total_score = 0
+    end
+
+    def total(x, y)
+    @score_card << [x, y]
+    if @score_card.length > 1
```
edpackard, line 15: this is good and useful logic (in the lines below), but it isn't immediately apparent what it is doing because of all the indexes (always a problem with arrays) - is there a possibility of maybe assigning some of these values to variables with clear names, or would that clutter things up too much?

lib/score.rb
```diff
@@ -0,0 +1,66 @@
...
+            @total_score += (x + y)
+        end
+    else 
+        @total_score += (x + y)
+    end
+    p "Your current total is #{@total_score}"
+    end
+
+    def strike(x, y)
+        #IF THREE IN A ROW, ADD THIRTY
```
edpackard, line 32: does this logic handle more than three in a row, or a perfect game?

edpackard (commented):
I'm impressed with the clarity of your control flow, both in the 'vanilla' method-based scorecard, and the class-based scorecard. It is straightforward to figure out how the various bits of your code interact with each other. You have also implemented a clear UI to deal with inputs, and provide messages to let the user know the result of their inputs. Your logic is also clear, albeit sometimes it gets a little blurry when arrays indexes come into the picture, although with a bit of extra thinking it is still clear what your code is doing.

There is certainly scope to build in checks for invalid inputs/edge cases, but that shouldn't be too tricky - especially in your case-based code.

You've thought about tests and have started writing them, but there aren't many finished tests here. However, on the whole, this is great and makes me feel like I massively overcomplicated the task - you've given me a lot of food for thought re: code clarity in any case :)

lib/score.rb
```diff
@@ -0,0 +1,66 @@
...
+        if @score_card.length >= 3 && @score_card[-1][0] == 10 && @score_card[-2][0] == 10 && @score_card[-3][0] == 10
```
eoinmakers, line 33: Perhaps drawing out a frame class could help with simplifying this logic a bit. 

It's hard to read this logic, because something like @score_card[-1][0] is difficult to understand if you don't have context from writing it - it's likely also difficult for the developer who wrote it to keep tabs on. 

If you instead used a Frame class to track the value of each roll, and then manipulated an array of Frames instead in order to tally up the score, it might make this slightly more readable, thus easier to refactor.

lib/score.rb
```diff
@@ -0,0 +1,66 @@
...
+            @total_score += 30
+        # IF TWO IN A ROW, ADD TWENTY
+        elsif
+            @score_card.length >= 2 && @score_card[-2][0] == 10 && @score_card[-1][0] == 10
```
eoinmakers, line 37: This could maybe be simplified a bit by drawing out these combinations of conditionals into methods that return booleans, and named something like 'is_three_in_a_row?', for example (not my best naming, but something like that). Descriptive methods that condense combinations of conditionals makes it easier for you to refactor logic like this, because you're thinking of descriptions of code / checks, rather than being faced with the logic itself, which isn't easy to understand just by reading.

lib/score.rb
```diff
@@ -0,0 +1,66 @@
...
+            @total_score += 20
+        # IF ONE PREVIOUSLY, ADD 10 AND X + Y
+        else 
+            @one_strike = ((x+y) + 10)
+            @total_score += @one_strike
+        end
```
eoinmakers, line 43: Perhaps a Switch statement instead of if/else would also make this more easily readable: https://stackoverflow.com/questions/427760/when-to-use-if-else-if-else-over-switch-statements-and-vice-versa

lib/score.rb
```diff
@@ -0,0 +1,66 @@
...
+    end    
+
+    def last_round(x,y,z)
+        # If the first frame is a strike, then it follows the usual strike pattern because there are two left
+        @score_card << [x]
+        if x == 10
+           total(x, 0)
+        end
+        # If the second frame is a strike, it doesn't follow the same pattern because you only get extra points
+        @score_card << [y]
```
eoinmakers, line 53: I think this is another place where refactoring to use a Frame class could make things more easily readable: if you could write frame.is_strike? in order to determine if it's a strike or not, you could keep the logic for this method just for calculating the score based on that evaluation, rather than having to do both in one place. 

```
def is_strike?
  # etc. . . 
end
```

lib/score.rb
```diff
@@ -0,0 +1,66 @@
...
+        else total(x,y)
+        end
+        if (x + y) == 10 || y == 10
+            @score_card << [z]
+            @total_score += z
+            p "Your total score is #{total_score}"
+        end
+    end
+
+end
```
eoinmakers, line 66: I really recommend the use of pseudocode for this challenge, because figuring out the logic, and then having to re-figure out the logic whenever you need to make a change or refactor anything, can be really confusing. 

https://www.unf.edu/~broggio/cop2221/2221pseu.htm

I would maybe write pseudocode for the base game logic first, and then try to write the pseudocode for my existing implementation, and then try to see if there are any improvements to be made in the logic just for the base game - and trying to draw that out first, eventually picking up one bonus game to try to add on, and to take it from there.

PR #140, opened 2021-08-31
bowling-challenge-ruby

PaulRainsford:


lib/bowling.rb
```diff
@@ -0,0 +1,48 @@
...
+    @rolls[current_roll + 1] + @rolls[current_roll + 2] + 10
+  end
+  
+  def spare?(current_roll)
+    @rolls[current_roll] + @rolls[current_roll + 1] == 10
+  end
+  
+  def strike?(current_roll)
+    @rolls[current_roll] == 10
+  end
```
Battery0, line 43: Unsure on this, but if current_roll can be either a first or second roll. Does that mean if the second roll == 10, then it's a strike? As i'm not 100% certain but a strike can only happen on the first roll of a frame. May need to look into splitting up the rolls into two different rolls per frame

lib/bowling.rb
```diff
@@ -0,0 +1,48 @@
+class Bowling
+  def initialize
+    @rolls = []
+  end
+  
+  def roll(pins)
+    raise 'You must enter a score less than 10' if pins > 10
```
Battery0, line 7: nice guard clause!

lib/bowling.rb
```diff
@@ -0,0 +1,48 @@
+class Bowling
```
Battery0, line 1: Might want to look into splitting this class into several classes so that you stick to Separation of Concerns

lib/bowling.rb
Battery0: I like the way you've accounted for spares & strikes and how to accrue bonuses!

lib/bowling.rb
Battery0: Not sure where else to add this comment so I will do it here. Might want to think about updating the README with your planning, how the game works, what the user must do to install everything to get it to work, etc, etc

spec/bowling_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'bowling'
+
```
Battery0, line 2: Couldn't see a good place to put this. I see in your code you have a guard clause, may want to right a test for it to see if it works as you intend it to?

PR #142, opened 2021-08-31
Weekend challenge: week five - incomplete

edpackard:
Basically complete, just need to hook up the scoring class to the frame and game classes (there is some commented out code in the frame class related to this - please ignore).

I enjoyed this fiendish challenge, but perhaps spent too much time refactoring and over-thinking frame class and game class tests. My scoring class is a bit rushed, but it does pass the tests - will hopefully get a chance to refactor in future (possibly when I'm hooking it up to the game / frame classes).

Would appreciate feedback on how I've tried to encapsulate my classes, especially frame. Also whether or not my game class tests are any good - I have attempted to test behaviour rather than state.


oddpotato:
Comments - 

Refactoring - 

If you're running validity checks on every throw, it is probably best to initialize a new instance of validity when the game starts, and then run methods from that class when a check is needed

The display at the end of every round is very pretty. I would separate that from the background logic. This is for readability but also because the bulk of the work is very complicated. You could also initialize a new scorecard display at the beginning of the game and append at the end of every round.

Great - 

There's a bunch of validation and tests and that's amazing long-term thinking.

Not So Great - 

This is very difficult for me to read coming in cold. Separating the basic functionality from the extra is very difficult. (This might be me)

Would Be Nice To Have - 

As the person inputting the data I'd like some feedback when I successfully add a score.

PR #171, opened 2021-11-01
For review

Kntanos:


game.rb
```diff
@@ -0,0 +1,38 @@
+class Game
+
+  require_relative 'score'
+
+  attr_reader :frames
+
+  def initialize
+    @frames = Array.new(10){ [] }
```
maxcfry, line 8: Interesting - I've never seen this kind of array syntax! Given that it's dealing with frames I'm assuming this is your way of setting the maximum amount of elements in the array as 10. But what I don't get is what ```{ [] }``` that is for?

game.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def initialize
+    @frames = Array.new(10){ [] }
+    @frame_number = 1
+    @roll_number = 1
+    @score = Score.new
+  end
+
+  def roll(pins)
+    if @frame_number < 10
```
maxcfry, line 15: It's interesting that you've got the main score calculated here - but that the strike score is calculated in the Score class

game.rb
```diff
@@ -0,0 +1,38 @@
...
+      else
+        @frames[@frame_number -1] << pins
+        @frame_number += 1
+      end
+    else
+      @frames[9] << pins
+    end
+  end
+
+  def score
```
maxcfry, line 32: There's a nice separation of concerns - you've got this adding up the total score - but other parts of the program computing all the complex stuff. A less elegant program would probably have all the score calculated in one method (and correspondingly - any minor tweaks to the program would have knock-on effects through the program). As far as I can tell - you've avoided some of those hazards.

score.rb
```diff
@@ -0,0 +1,35 @@
+class Score
+
+  def game_total(frames)
+    add_bonuses(frames)
+    frames.flatten.compact.sum
```
maxcfry, line 5: This is cool - I've never seen the .compact word!

maxcfry: Just read up about it - that is such an elegant way to account for gutter balls (just set those scores to nil, then eliminate them from the score total using compact.)

PR #176, opened 2021-11-21
Tatiana - Bowling Challenge - Week5

tatiananantes:


lib/game.rb
```diff
@@ -0,0 +1,43 @@
+class Game
```
Zimmja, line 1: Well done for completing the challenge! Your code works! You got the core logic done and have full rspec coverage

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+  def initialize
+    @rolls = [] 
+  end
+
+  def roll(pins)
```
Zimmja, line 6: I've noticed this code only works for full games, which is fine but it does mean it's quite hard to test in irb, you have to enter every roll one-after-another. I'd suggest making your life easier by having some sort of function that can be called on to input a list of rolls immediately, something like

def test
[1, 4, 4, 5, 6, 4, 5, 5, 10, 0, 1, 7, 3, 6, 4, 10, 2, 8, 6].each { |my_roll| mygame.roll(my_roll) }
end

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+        rollIndex += 2
+      else
+        frame_points += @rolls[rollIndex] + @rolls[rollIndex + 1]
+        rollIndex += 2
+      end
+    end
+    frame_points
+  end
+
+def isSpare?(rollIndex)
```
Zimmja, line 29: I think this is really clever. I didn't recognise that single values could be used ot calculate a spare / strike bonus, it's a great spot from you and well executed

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+  @rolls[rollIndex] + @rolls[rollIndex + 1] == 10
+end
+
+def sparePoints(rollIndex)
```
Zimmja, line 33: Lovely clean one-liner methods

lib/game.rb
```diff
@@ -0,0 +1,43 @@
+class Game
+  def initialize
+    @rolls = [] 
+  end
+
+  def roll(pins)
+    @rolls.push(pins)
+  end
+
+  def points
```
Zimmja, line 10: As far as I can tell, this seems to work perfectly. You can use the roll method to add too many rolls to the scorecard, but even so this method only counts rolls within a functional game and it seems to calculate correctly. I've tried to break it and I can't, great job

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+    frame_points = 0
+    rollIndex = 0
+    # calculates points per game frames(ten)
+    10.times do
+      if isStrike?(rollIndex)
```
Zimmja, line 15: The more I look at this the cleverer I think it is, amazing job!

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+      elsif isSpare?(rollIndex)
+        frame_points += sparePoints(rollIndex)
+        rollIndex += 2
+      else
+        frame_points += @rolls[rollIndex] + @rolls[rollIndex + 1]
+        rollIndex += 2
+      end
+    end
+    frame_points
+  end
```
Zimmja, line 27: The points method is quite long. I think I've read somewhere that methods should ideally be no longer than 2 or 3 lines. Is there  a way to break this method down into smaller methods?

spec/game_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'game'
+
+describe Game do
+  
+  it 'calculates a gutter game' do
+    game = Game.new
```
Zimmja, line 6: You don't need to say "game = Game.new" at the beginning of each statement. You can either declare this once at the top by saying "let(:game) { Game.new}", or because you're not initiating with any arguments, you can just use the word "subject" instead of "game" as in "20.times { subject.roll(0) }"

spec/game_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'game'
+
+describe Game do
+  
+  it 'calculates a gutter game' do
```
Zimmja, line 5: This is specifically describing the results of the "points" method, therefore I think you should incude this in another description block, e.g.
`describe "#points" do`
`it "calculates a gutter game"`

PR #177, opened 2021-11-21
Ginny - bowling scorecard challenge

ginnyamazed:
Partial progress.  Was sick over the weekend and unable to offer full attention.  Also ran into rspec issues.  
If I was able to focus and keep working I would have continued to implement frame count and scoring methods or pulling these methods into the game class.  I would also have tried to figure out how to represent strikes and spares in code.


lib/score.rb
```diff
@@ -0,0 +1,19 @@
+# score class
```
bmargerison, line 1: your structure is nice, good choice of classes

lib/score.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def frame_total
+    @frame_total << @t1p
+    @frame_total << @t2p
+    @frame_total.sum
+  end
+
+  def strike_score
+    @frame_total = 10 + @st1 + @st2
```
bmargerison, line 17: this is the difficult bit! trying to get the real time score for a strike when the next rolls haven't been done yet

lib/frame.rb
```diff
@@ -0,0 +1,42 @@
...
+    elsif @t1p = 10
+      strike
+    else puts 'Please enter your pin count again: '
+    end
+  end
+
+  def turn2
+    puts 'Please enter the number of pins you knocked down on turn 1: '
+    @t2p = $stdin.gets.chomp.to_i
+    if @t2p <=9 && (@t1p + @t2p <=9)
```
bmargerison, line 25: nice succinct logic

spec/frame_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+# require 'frame'
+#
+# describe Frame do
+#
+#   frame = Frame.new
+#
+#   describe 'turn1' do
+#     it 'accepts user input for their first turn count of pins' do
+#       allow($stdin).to receive(:gets).and_return('5')
```
bmargerison, line 10: trying to stub the user input, very good!

PR #185, opened 2021-11-22
ben-m bowling-challenge

bmargerison:
I have managed to get the logic working for the game, but since this was such a brain bender I didn't really test drive the development. This has probably led to some very messy and superfluous code. I think the classes are laid out well, but some refactoring will be required, especially of the Scorecard class. Also additional work is required on the tests.


lib/frame.rb
```diff
@@ -0,0 +1,50 @@
+require './lib/roll'
```
ginnyamazed, line 1: great use of self contained classes and single purpose.

PR #195, opened 2022-01-04
My attempt at the bowling weekend challenge

chris-clement:


spec/bowling_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+      @bowling.input(2)
+      @bowling.input(7)
+      expect(@bowling.score).to eq [3, 2, 7]
+    end
+  end
+
+  describe '#last_frame' do
+    it 
+  end
+end
```
bryn9696, line 134: Clear and concise written tests where I can see what they're testing for and grouped together well

spec/bowling_spec.rb
```diff
@@ -0,0 +1,134 @@
+# frozen_string_literal: true
+
+require 'bowling'
+
+describe Bowling do
+  before(:each) do
+    @bowling = Bowling.new
+  end
+
```
bryn9696, line 9: Good use of a before statement

lib/bowling.rb
```diff
@@ -0,0 +1,99 @@
...
+      record_strike
+    end
+    bonus_points(pins)
+    add_to_score(pins)
+    add_to_total_score(pins)
+    increase_frame_or_turn
+  end
+
+  private
+
```
bryn9696, line 28: Good use of public and private

lib/bowling.rb
```diff
@@ -0,0 +1,99 @@
...
+  def spare?
+    if @frame == 1
+      false
+    elsif @score[-1] == 10
+      false
+    else
+      @score[-1] + @score[-2] == 10
+    end
+  end
+
```
bryn9696, line 75: Great logic here for whether its a spare or not

lib/bowling.rb
```diff
@@ -0,0 +1,99 @@
...
+  def bonus_points(pins)
+    if spare? && @turn == 1
+      add_to_total_score(pins)
+    elsif check_strike(@frame - 1) && @turn == 2
+      add_to_total_score(pins + @score[-1])
+    end
+  end
+
```
bryn9696, line 95: efficient way to implement the bonus points

bryn9696 (commented):
.

lib/bowling.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+class Bowling
+  attr_reader :turn, :frame, :total_score, :score, :strike_frames
```
eoinmakers, line 4: Do all of these variables need to be exposed outside of the class?

eoinmakers (commented):
Good work so far. Your approach in starting with the basic rules was good, and you've tested for gutter games, spares and strikes. 

One area that could be a focus of a refactor would be that you currently have one monolithic class that handles all of the logic. This means that unfurling this logic to make changes to individual aspects of the game can be challenging. 

An approach in dividing them could look something like: 
- knowing whether a frame is a strike / spare
- keeping track of the progress of a game
- calculating bonuses

PR #196, opened 2022-01-04
Bowling-Challenge-Ruby

TomasGarciaDev:
Nasty code, lots of refactoring to be done


lib/bowling.rb
```diff
@@ -0,0 +1,110 @@
+
+class Bowling
+
+  attr_reader :frames, :frame, :spare_counter, :strike_counter, :strike_counter2
+  
+  def initialize
+    @frames = []
+    @frame = []
```
kim-morgan, line 8: Nice clear naming of arrays

lib/bowling.rb
```diff
@@ -0,0 +1,110 @@
...
+    @spare_counter = []
+    @strike_counter = []
+    @strike_counter2 = []
+  end
+  
+  def roll(pins)
```
kim-morgan, line 14: Okay lol so my first comment would be that you could split each of these if statements into their own method, preferably with a name describing what exactly it does?

I can see that you were really thorough about accounting for every possibility, but as an outsider it's quite hard to work out what each of the if statements actually means. So you could split each of them into their own method with a name like strike_followed_by_spare etc and it would be a lot easier to follow and also be more in line with SRP.

I also think that it would be a good idea to totally separate rolls and calculating bonuses. When I first started doing it I also had them together but it just ended up getting so complicated... so maybe have the rolls be input, and then have calculating bonuses as something totally separate based on the rolls/frames you have. Just an idea. I think that will help you cut down on all the different if/elsif!

lib/bowling.rb
```diff
@@ -0,0 +1,110 @@
...
+      elsif @frame.length == 2
+        @frames << @frame
+        @frame = []
+      end
+    end
+  end
+
+  def score 
+    score = 0
+    @frames.each do |frame|
```
kim-morgan, line 94: Alternative suggestion:
@frames.map { |frame| frame.sum }.sum

lib/bowling.rb
```diff
@@ -0,0 +1,110 @@
...
+       score += frame.sum
+    end
+    score
+  end 
+
+  private
```
kim-morgan, line 100: Nice use of private methods

lib/bowling.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  def spare?(frame)
+    frame.length == 2 && frame.sum == 10
```
kim-morgan, line 103: Concise and simple to understand

lib/bowling.rb
```diff
@@ -0,0 +1,110 @@
...
+  end 
+
+  private
+
+  def spare?(frame)
+    frame.length == 2 && frame.sum == 10 
+  end
+
+  def strike?(frame)
+    frame == [10]
```
kim-morgan, line 107: Concise and simple to understand

spec/bowling_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    @game.roll(1)
+    @game.roll(7)
+    @game.roll(3)
+    @game.roll(6)
+    @game.roll(4)
+    @game.roll(10)
+    @game.roll(2)
+    @game.roll(8)
+    @game.roll(6)
+    expect(@game.score).to eq 133
```
kim-morgan, line 116: Good job testing a variety of different cases

PR #197, opened 2022-01-04
Bowling scorecard weekend challenge

FrancescoCollotto:


lib/scoreboard.rb
```diff
@@ -0,0 +1,42 @@
...
+      if strike?(index)
+        score += 10 + strike_bonus_points(index)
+      elsif spare?(index)
+        score += 10 + spare_bonus_points(index)
+      else
+        score += regular_points(index)
+      end
+      strike?(index) ? index += 1 : index += 2
+    end
+    score
```
jessgordon, line 19: Honestly your code is so DRY and a pleasure to read that I really am being picky here to try and find something to comment on! 
Perhaps using a different variable name to the name of the method (score) might keep this code a bit more readable and avoid accidentally calling the score method instead of returning the score variable here - just thinking in some languages this might recursively call the score method?

lib/scoreboard.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def strike?(index)
+    true if @rolls[index] == 10
```
jessgordon, line 23: may also work without the 'true if' statement and return true/ false

lib/scoreboard.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def spare?(index)
+    true if @rolls[index] + @rolls[index + 1] == 10
+  end
```
jessgordon, line 28: may also work without the 'true if' statement and return true/ false

jessgordon (commented):
It was very difficult to find anything to comment on here - your code is so easy to read, TDD tests all situations, code is DRY, really nice work! 

In terms of the goals for the challenge: 
Using diagramming to plan your approach to the challenge - unknown
TDD your code - 10/10 so readable and tests all situations
Focus on testing behaviour rather than state - you test each method used to produce the final score
Commit often, with good commit messages - commits are clear
Single Responsibility Principle and encapsulation - methods adhere to SRP, can't see the need for more than one class with your code - that would just make it unnecessarily complicated!
Clear and readable code - 10/10 pleasure to read

PR #198, opened 2022-01-04
Bowling challenge

kim-morgan:
Hi reviewer!
I didn't manage to write a readme, but basically there are two classes:

- Game: This controls the flow of the game and scoring. Users input rolls using the "input_roll" method, which records them and starts a new frame if necessary
- Frame: Stores info about each frame

It should be able to calculate the score accurately of any game, including games with extra rolls and bonuses in the last round. The tests check several normal games as well as special cases such as perfect game and they score correctly.

One thing that doesn't sit well with me, I suppose, is that my method for calculating score bonuses is quite complicated...


TomasGarciaDev (commented):
Great clean code, really nice that you split the logic of the frames and the game into 2 classes. 
Your methods are impressively short, well done!!
I think for a fully functional Model to be used in a UI it would be nice to have the individual score of each frame, and the score at the point of each frame.

PR #200, opened 2022-01-04
Bowling Challenge

jessgordon:


FrancescoCollotto (commented):
Really nice work, you actually implemented a proper bowling game, good separation of classes, nice written tests.
It is not easy to follow your code, it seems over complicated
overall very good job! well done!!

PR #201, opened 2022-01-04
Bowling Challenge

msc49:


lib/game.rb
```diff
@@ -0,0 +1,29 @@
+require 'bowl'
+
+class Game
```
zidehen, line 3: nice separation of classes

spec/bowl_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'bowl'
+
```
zidehen, line 2: It's good that you have very thorough tests

spec/bowl_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  it 'can accurately calculate spare' do
+    bowl.first_roll(5)
+    bowl.second_roll(5)
+    bowl.first_roll(3)
+    expect(bowl.rolls[0][:total_roll]).to eq 13
+
+  end
+
+  it 'can accurately calculate spare' do
```
zidehen, line 58: could name this test different from the test before to prevent any confusion

lib/bowl.rb
```diff
@@ -0,0 +1,77 @@
...
+
+
+  def total_score
+    @rolls.sum{|hash| hash[:total_roll]}
+  end
+
+
+
+  
+  private
```
zidehen, line 44: well done for refactoring into private methods - very clean

lib/bowl.rb
```diff
@@ -0,0 +1,77 @@
+class Bowl
+
+  attr_reader :rolls
+
+  def initialize
+    @rolls = [
```
zidehen, line 6: I didn't think to use hashes for this so well done for implementing this

zidehen (commented):
Great work!

PR #205, opened 2022-01-07
finished bowling

KKassim:


spec/bowling_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'bowling'
+
+describe BowlingGame do
+
+  before do
+    @game = BowlingGame.new
+  end
+ 
+  it "can play a game of gutter balls" do
```
s-dousse, line 9: clear tests! nice

lib/bowling.rb
```diff
@@ -0,0 +1,44 @@
...
+    @rolls = []
+  end
+
+  def rolls(pins)
+    @rolls.push(pins)
+  end
+    
+  def score
+    totalScore = 0
+    rollTurn = 0
```
s-dousse, line 13: Just a suggestion: a bit of repetition here; when passing rollTurn as an argument for each function, could it be replaced by the index of your @rolls?

PR #206, opened 2022-01-23
Bowling_Challenge

robinucar:


README.md
```diff
@@ -1,3 +1,19 @@
+My Approach
```
kat-mcbrooks, line 1: So good to see your approach listed like this.

lib/scoreboard.rb
```diff
@@ -27,28 +27,28 @@ class Scoreboard
...
-				score += frames[index + 1][0..1].sum # it will add next frame score to the strike score which is 10
-			elsif frame.sum == 10 # if frame is spare
-				score += frames[index + 1][0] # it will add next frame first rolle score to the spare frame score which is 10
-			end
-		end
-		score # return the score
+    frames.each_with_index do |frame, index|
+      score += frame.sum # this will add the sum of rolls in each frame to the score
+      if frame [0] == 10 # if frame is strike
+        score += frames[index + 1][0..1].sum # it will add next frame score to the strike score which is 10
```
kat-mcbrooks, line 47: Have you tried a test with a scoreboard where there are 2 strikes in a row? e.g. [10,0], [10,0], [8,0]. I believe the bonus points for the first strike frame should be the sum of the next two frames i.e. in this example that would be 18 so the total points for the first frame in this example would be 28 (10+18 bonus). However I think your program would calculate the bonus to be 10 . Please do tell me if I've misunderstood though!

kat-mcbrooks (commented):
You solved this so cleanly! Excellent code and clear that you carefully TDD-ed your approach

lib/scoreboard.rb
```diff
@@ -0,0 +1,54 @@
...
+    [10],
+    [10],
+    [10],
+    [10],
+    [10],
+    [10],
+    [10, 10, 10]
+  ].freeze
+  def initialize(frames)
+    @result = case frames
```
alicelieutier, line 32: Your class here is not really used as a class. It calculates and stores a result at initialisation, so it basically is just acting as a function.
Have you modelled what your program's input/output should be? In other words, how would you like the user to use your programme? What will they call, in what order?

spec/scoreboard_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      expect(scoreboard.result).to eq 0
+    end
+    it 'returns maximum score of game as 300' do
+      expect(scoreboard_1.result).to eq 300
+    end
+  end
+  describe '#calculate_score' do
+    it 'should calculate_score' do
+      expect(scoreboard_1).to respond_to(:calculate_score)
+    end
```
alicelieutier, line 57: It seems like calculate score acts as a private or helper function in your class, not really a function that should be called directly by the user. In fact, in all the following tests you are testing the result, not calling calculate_scores directly.
I think you should remove this test and put all the other ones in the `describe .result` as they are using `.result`

lib/scoreboard.rb
alicelieutier: I agree with @kat-mcbrooks - there's a bug here :)

lib/scoreboard.rb
alicelieutier: Have you tested with games that have strikes or spares in the tenth frame? I think your data model doesn't work in those circumstances. If there's a strike or a spare in the tenth frame, the player needs to roll more balls in order to calculate bonuses correctly, but these rolls do not count as a frame by themselves, only as bonuses.
In your data model, there's no good way to represent these additional rolls, as you are passing rolls that represent frames directly.

One option you could move to is to store an array of rolls instead of frames, letting the logic of your code decides which frames the rolls should belong to if any.

PR #210, opened 2022-01-24
Kiran's Bowling Challenge Ruby

KiranKukar:


lib/frames.rb
```diff
@@ -0,0 +1,77 @@
...
+  def new_round
+    if @round >= 11
+      ScoreCard.new(@score_count, @strike_count)
+    else
+      new_roll
+    end
+  end
+
+  def new_roll
+    @roll = Roll.new(score_count: @score_count, strike_count: @strike_count, first_roll: @first_roll, this_frame_count: @this_frame_count)
```
JohnForster, line 30: Each roll currently needs to know a lot about what's going on in the rest of the game, which suggests that maybe some responsibilities could be moved around. Ideally, a roll should be pretty independent of the rest of the game.

JohnForster, line 30: Each roll currently needs to know a lot about what's going on in the rest of the game, which suggests that maybe some responsibilities could be moved out of it, or the class should be renamed. Ideally, a roll should be pretty independent of the rest of the game.

lib/frames.rb
```diff
@@ -0,0 +1,77 @@
+require './lib/score_card'
+require './lib/roll'
+
+class Frames
```
JohnForster, line 4: You normally want to keep class names singular, as plural indicates a list. Frame doesn't quite describe what this class is doing, can you think of an alternative singular name?

lib/roll.rb
```diff
@@ -0,0 +1,39 @@
+require './lib/frames'
+
+class Roll
```
JohnForster, line 3: As mentioned above, this class currently needs a lot of information about the rest of the game, suggesting that it isn't quite describing a single Roll. A single roll probably shouldn't care too much about the game it's been rolled in.  Can you think of a more accurately descriptive name for this class? 

If not, maybe you need to think a bit about what its responsibilities are. Does it make sense for them to be grouped together like this?

lib/frames.rb
```diff
@@ -0,0 +1,77 @@
...
+  def new_roll
+    @roll = Roll.new(score_count: @score_count, strike_count: @strike_count, first_roll: @first_roll, this_frame_count: @this_frame_count)
+    @roll.scoring(pins_knocked)
+    @bonus = @roll.strike_or_spare
+    roll_again?
+  end
+
+  def roll_again?
+    @roll.next_roll?
+    if @roll.roll_again == true
```
JohnForster, line 38: Where possible, it's better to use return statements to communicate, rather than setting and reading instance variables. For example, this could read

```ruby
  roll_again = @roll.next_roll?
  if roll_again 
    @first_roll = false
    new_roll
  else ...
```

Ideally, you want to be minimising the state of a class as much as possible, and using return values instead.

JohnForster (commented):
There's a lot of great code here, your testing especially is very thorough, and has good coverage.

It might be an idea to step back and have a think about how the problem could be broken down. A Game is made up of many Frames, a Frame is made up of a number of Rolls etc. This can help you to structure your code, as Game, Frame and Roll could be their own classes. It then allows you to think about the responsibilities of each class more clearly.

eg. Which class calculates the overall score? Which class calculates the bonus from a spare or strike?

Separating the communication to the user into the Scorecard class is a nice touch. Could the Scorecard have a Game associated with it? Or might a Game have a Scorecard?

This is a strong first attempt, but having a think about how to break the problem down in advance (maybe draw out a Domain Model Diagram?) could make this even better.

lib/frames.rb
JohnForster: Can you spot any other places where you could return a value instead of storing it in an instance variable?

PR #211, opened 2022-01-24
Bowling Scorecard Challenge - Alex

Alex-Jamshidi:
Working, not refactored or fully tested.


mikitaro-27 (commented):
[This may not be right]

I found it difficult to review the code as the UI gets in the way of understanding what the classes and functions are actually doing. There seems to be too many things that each of them are supposed to be doing. Maybe a good refactor would be to take UI out of the logic?

We did one challenge with student_directory a while ago where we separated the UI quite neatly. You could have a look at that.

The UI also affect the readability of the tests as well as you need to stub out the gets. On top, you are mostly testing for the string output. I think that separating the logic from UI would allow you to have more descriptive tests. 

The script seems to be working and doing its job so you should be able to extract the logic from it and refactor everything nicely, keeping in mind single responsibility principle.

PR #215, opened 2022-01-24
Bowling 

ElliottZaki:
ScoreBoard not fully implemented. Also wanted to separate score into another class.


cris-ch:
Great code that works!
Maybe trying to have one or two more classes like is says in the last commit would have been a good idea.

PR #218, opened 2022-01-24
Bowling Challenge - Fran

fpmrqs:


README.md
```diff
@@ -1,27 +1,18 @@
 Bowling Challenge in Ruby
 =================
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday week
+* This is the weekend challenge for week 5 of Makers Academy.
```
hannahdesmond, line 4: I like that you have commented here your potential plans for an interface and that you can run it in irb. It might also be nice to see any diagramming if you did that 🙂

spec/bowling_scorecard_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'bowling_scorecard.rb'
+
+describe BowlingScorecard do
+  describe '.add_knocked_pins' do
+    it 'registers number of pins knocked in a roll' do
```
hannahdesmond, line 5: If you were refactoring, I wonder if you could get rid of this test because your third test tests this method.

lib/frame.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  attr_accessor :frames
+  def initialize
+    @frames = []
+  end
+
+  def create_new(rolls)
+    add_bonus_points(rolls)
+
+    if !tenth_frame?
```
hannahdesmond, line 11: You could refactor this into a ternary operator.

hannahdesmond (commented):
Your code is really smart Fran. I was having a look back at various old Makers’ repos last night and I think yours is the shortest one of all of them – it’s great! It’s easy to read, it’s clear what your variables and methods do. It’s clean and your methods don’t do too much so it looks like you’ve adhered well to SRP. Well done.
Re: tests, I haven’t gone through those in as much detail but they look fairly comprehensive from what I can tell. You covered some edge cases like being given more than 10 pins for a frame.

PR #219, opened 2022-01-24
bowling challenge incomplete 

naima90:


lib/bowling.rb
```diff
@@ -1,16 +1,25 @@
 class Bowling
```
BenCoyne, line 1: Great start to the project! :)

spec/bowling_spec.rb
```diff
@@ -8,9 +8,22 @@
     expect(bowling.score).to eq 10
```
BenCoyne, line 8: Nice testing here! You could consider wrapping your tests for #roll within its own describe block, though this is a matter of preference really. All in all though great so far!

BenCoyne (commented):
Thank you for letting me review your code !

PR #221, opened 2022-01-24
bowling-challenge-ruby

cmkap:
weekend challenge


lib/frame.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+  
+    def strike?
+      if (@contents.sum == 10) && (@contents.length == 1)
+        true
+      else
+        false
+      end
+    end
+  end
```
icvotria, line 35: Really nice and clean!! For the predicate methods spare? and strike?, I think you can just make the statement, and the method will return true or false without it needing to be in an if statement, eg. 
def spare?
  (@contents.sum == 10) && (@contents.length == 2)
end
That should return true if it's true and false if not, without needing the if... else.

lib/score.rb
```diff
@@ -0,0 +1,68 @@
...
+  def initialize
+    @completed_frames = []
+    @current_frame = Frame.new
+  end
+
+  def check_current_frame
+    if @current_frame.rolls == 2 || @current_frame.strike? || @completed_frames.length >= 10
+      @completed_frames << @current_frame.contents
+      @current_frame = Frame.new
+    end
```
icvotria, line 15: this is really clever

lib/score.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def strike?(frame)
+    if (frame.sum == 10) && (frame.length == 1)
+      true
+    else
+      false
+    end
+  end
+
```
icvotria, line 58: Did you try using the spare? and strike? methods from the frame class here? I feel like there must be a way to just use one set of them, either in here or in the other class, but I can't quite think of how you'd do it

lib/score.rb
```diff
@@ -0,0 +1,68 @@
...
+  def apply_bonuses(frames = @completed_frames)
+    frames.each_with_index do |frame, index|
+      if spare?(frame) && (frames[index + 1])
+        frame[-1] += frames[index + 1][0]
+      elsif strike?(frame) && (frames[index + 1]) && (frames[index + 1].length == 2)
+        frame[-1] += (frames[index + 1][0] + frames[index + 1][1])
+      elsif strike?(frame) && (frames[index + 1]) && (frames[index + 2])
+        frame[-1] += (frames[index + 1][0] + frames[index + 2][0])
+      end
+    end
```
icvotria, line 38: Did you try using the spare? and strike? methods from the frame class in here rather than making new ones? I feel like there must be a way to just use one set of them, either in here or in the other class. Maybe if you passed @current_frame.contents into the methods below as an argument?

lib/score.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def bonus_roll_correction(array)
+    array.delete_if.with_index { |_num, idx| idx > 9 }
+  end
+
+  def roll_limiter
+    if (@completed_frames.length == 10 && @completed_frames[-1].sum != 10) || @completed_frames.length >= 12
+      raise "That was the last frame, kid. You want to roll again, start a new game"
+    end
```
icvotria, line 66: excellent error message

spec/features_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+    score.record_roll(2)
+    score.record_roll(2)
+    score.record_roll(2)
+    score.record_roll(2)
+    score.record_roll(2)
+    score.record_roll(2)
+    score.record_roll(2)
+    score.record_roll(2)
+    score.record_roll(2)
+    score.record_roll(2)
```
icvotria, line 96: you can do something like 18.times { score.record_roll(2) } (I'm not sure that's quite right but something like that) to save yourself a bunch of lines :)

spec/features_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+    score.record_roll(8)
+    score.record_roll(10)
+    expect(score.calculate_score).to eq(129)
+    expect{score.record_roll(2)}.to raise_error("That was the last frame, kid. You want to roll again, start a new game")
+  end
+
+  it 'does not allow more than two bonus rolls if a strike is rolled' do
+    expect{ 13.times{score.record_roll(10)} }.to raise_error("That was the last frame, kid. You want to roll again, start a new game")
+  end
+end
```
icvotria, line 207: Really thorough testing, I feel like you covered all the weird little aspects of bowling that might cause hiccups and stuff, plus edge cases

spec/score_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      expect(score.completed_frames).to eq([[2, 7]])
+    end
+    it 'empties @current_frame into @completed_frames if it contains a strike' do
+      allow(score).to receive(:current_frame).and_return(frame)
+      allow(frame).to receive(:strike?).and_return(true)
+
+      expect(score.check_current_frame).to eq([])
+    end
+  end
+end
```
icvotria, line 45: good mocking :)

icvotria (commented):
Really good overall! Super clean and tidy and easy to understand, nice work :)

PR #227, opened 2022-01-24
Bowling challenge (logic) - kat

kat-mcbrooks:
Grateful of feedback on overall approach. Also had these particular questions/uncertainties:

1. I hope I've interpreted the challenge instructions correctly - instructions did not seem to specify a requirement that a player can input rolls one by one, so I have assumed the scorecard can be entered in its entirety. Or should the scorecard allow players to enter an individual frame roll?!
2. I did debate about creating a Frame class, but couldn't think of how to 'inject' a total frames array into the Scorecard class so left it as is (see frame.rb for early thoughts)
3. Not sure about my tests - would appreciate feedback. I believe/hope I'm testing behaviour over state...?! Also I think I've ended up with redundant tests and wondered if perhaps the only test I needed was to check that total_score is calculated correctly for various potential score inputs


penguat:
I believe you have interpreted the instructions correctly, there's certainly nothing in there to say that rolls must be input one by one. It's typical when bowling that your pins knocked down would be recorded after each ball bowled, so your choice is atypical but not wrong.

PR #228, opened 2022-01-30
Bowling-Challenge-Ruby

hannahdesmond:


lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+  def initialize
+    @rolls = [] 
+    @max_rolls = 20
+  end
+
+  def roll(pins) 
+    check_pin_number
+    check_game_status
+    @rolls << pins
+      if @rolls.count == 19 && (pins == 10) || @rolls.count == 20 && (@rolls[-1] + @rolls[-2]) == 10
```
alicelieutier, line 15: You could extract a method called "strike on tenth frame" and one called "spare on tenth frame" here. It wouldn't save lines of code, but could make this a little bit easier to read.

lib/game.rb
```diff
@@ -0,0 +1,90 @@
+class Game
+  attr_reader :rolls
+
+  MAX_PINS = 10
+
+  def initialize
+    @rolls = [] 
+    @max_rolls = 20
```
alicelieutier, line 8: This is a pretty smart way to track additional bonuses!

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def total_score
+    pins_score + total_bonus_points
+  end
+  
+  private
+
+  def check_pin_number
+    raise ("This is not valid input.") if pins > MAX_PINS
```
alicelieutier, line 33: The error could be a bit more explicit here, to help the user. Here the user uses the code directly so they probably know what "input" they gave, but this class could be integrated in a bigger system. It's easier for the programmer to track down a bug if the error is more explicit. Something like: "A number between 0 and 10 was expected for the rolls, you entered X".


Which actually makes me realise - if you are going to check the number, you can also check it's positive :)

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+      if @rolls.count == 19 && (pins == 10) || @rolls.count == 20 && (@rolls[-1] + @rolls[-2]) == 10
+        @max_rolls += 1
+      elsif first_bowl? && (pins == 10)
+        @rolls << "-" unless completed_all_rolls
+      end
+  end
+
+  def check_game_status
+    raise ("Game Over. Ask for your score!") if completed_all_rolls
+  end
```
alicelieutier, line 24: This is an error, but the language is one of a normal user interaction. Do you expect the user to get this message at some point or should they avoid it?
You can explain the error more to them:
"Cannot roll after the game has completed. Maybe you wanted to ask for you score?"

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+    check_game_status
+    @rolls << pins
+      if @rolls.count == 19 && (pins == 10) || @rolls.count == 20 && (@rolls[-1] + @rolls[-2]) == 10
+        @max_rolls += 1
+      elsif first_bowl? && (pins == 10)
+        @rolls << "-" unless completed_all_rolls
+      end
+  end
+
+  def check_game_status
```
alicelieutier, line 22: This guard clause should probably be private. Also, the name is not super explicit, maybe "check_for_completed_game" or something like that?

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+  def check_pin_number
+    raise ("This is not valid input.") if pins > MAX_PINS
+  end
+
+  def completed_all_rolls
+    @rolls.count == @max_rolls
+  end
+
+  def pins_score 
+    @scores = @rolls - ["-"]
```
alicelieutier, line 41: Why are you introducing another instance variable here? A local variable is alway preferible.

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+          sum += frames[i][0] 
+        end
+      end
+    end
+    sum
+  end
+
+  def zero_pad
+    2.times { @rolls << 0 }
+  end
```
alicelieutier, line 69: This method changes the state of your system, even though it's part of the score, which should just provide the result of a calculation based on the state of your system.
This creates bugs, for example if you ask for total score before you have finished your game, now you have 4 empty rolls.

An easy way around could be that this method creates a copy of the rolls with added zeroes and returns it for the others to use. You could rename it to `zero_padded_rolls`.

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+    raise ("This is not valid input.") if pins > MAX_PINS
+  end
+
+  def completed_all_rolls
+    @rolls.count == @max_rolls
+  end
+
+  def pins_score 
+    @scores = @rolls - ["-"]
+    @scores.inject(0, :+)
```
alicelieutier, line 42: I believe ruby has a sum method.

lib/game.rb
alicelieutier: Would would then need to pass that as a argument to the frames method.

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+  def strike_score
+    sum = 0
+    zero_pad
+    frames.each_with_index do |frame, i| 
+      if frame.include?("-") 
+        i += 1
+        if frames[i][1] == "-" || frames[i][1] = 10
+          sum += frames[i][0]
+          i += 1
+          sum += frames[i][0]
```
alicelieutier, line 84: Does this work for one strike with normal rolls after AND for a perfect game with many strikes in a row?

spec/game_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  it 'only allows scores of 0-10' do
+    expect { game.roll(12) }.to raise_error("This is not valid input.")
+  end
+  it 'scores a spare' do
+    game.roll(5)
+    game.roll(5)
+    game.roll(5)
+    17.times { game.roll(0) }
+    expect(game.total_score).to eq(20)
+  end
```
alicelieutier, line 26: Missing a test for scoring just one strike in the middle of the game (or at the start). I think your programme doesn't work for that scenario.

alicelieutier (commented):
Great job! 
You have a bug and there's a few things you can refactor, but your code is well organised and clear, and your tests are good. 
Now to do this in JS ;)

PR #229, opened 2022-02-26
Bowling Challenge - Complete

nyahehnagi:
I'd like to improve on how I calculate bonus points. Pretty sure it can be further refactored..  I am still working on it

I've incorporated a LastFrame as an inherited class from Frame. This was for 2 reasons, I wanted to see how `super ` works and this seemed a good opportunity to see it in action, plus the last frame in bowling exhibed additional behaviour over a normal frame. Now.. I've read more.. it seems I can modules.. ok.. off I go to read about this and see if there's a refactor needed :)

I have intentionally not mocked the Frame and LastFrame classes in Scorecard. Although I have passed in these dependancies into Scorecard, when it comes to functionality I feel these classes are always going to be together. I cannot see how the Frame class could be used elsewhere but the Scorecard.  Adding mocks would a) be very time consuming for little value.. the tests for the Scorecard cover mulitple complex scenarios where the input from the Frame classes is key to score evaluation b) adding mocks would inhibit the refactoring process on Scorecard. I'd be keen to discuss this with a coach on their thoughts.. it could be my design is well off and to complex.. and that could be a warning sign to the design approach I took

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,19 @@ This is a program which calculates the score of a game of 10 pin bowling
 The following API is exposed from the ScoreCard class
 
 * current_frame_number - shows the current frame
-* log_roll - logs a roll from the player
-* score - shows the current score
+* log_roll - logs a roll from the player, pass in a number from 0-10
+* score - shows the current score at any point in the game
 
+----------------------
+## Installation
+----------------------
+Install gems with [bundler](https://bundler.io/ "bundler") 
+~~~~
+$> bundle install 
+~~~~
+To install code from the latest source
+~~~~
+git@github.com:nyahehnagi/bowling-challenge-ruby.git
 
 ## Testing
 -----------------------
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@ $> bundle install
 ~~~~
 To install code from the latest source
 ~~~~
-git@github.com:nyahehnagi/bowling-challenge-ruby.git
+git clone git@github.com:nyahehnagi/bowling-challenge-ruby.git
+~~~~
 
 ## Testing
 -----------------------
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ This is a program which calculates the score of a game of 10 pin bowling
 
 The following API is exposed from the ScoreCard class
 
-* current_frame_number - shows the current frame
-* log_roll - logs a roll from the player, pass in a number from 0-10
-* score - shows the current score at any point in the game
+* `#current_frame_number` - shows the current frame
+* `#log_roll(pins_downed)` - logs a roll from the player, pass in a number from 0-10
+* `#score` - shows the current score at any point in the game
 
 ----------------------
 ## Installation
--- a/README.md
+++ b/README.md
@@ -1,65 +1,17 @@
-Bowling Challenge in Ruby
-=================
+# Bowling Challenge
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday week
+This is a program which calculates the score of a game of 10 pin bowling
 
-## The Task
+The following API is exposed from the ScoreCard class
 
-**THIS IS NOT A BOWLING GAME, IT IS A BOWLING SCORECARD PROGRAM. DO NOT GENERATE RANDOM ROLLS. THE USER INPUTS THE ROLLS.**
+* current_frame_number - shows the current frame
+* log_roll - logs a roll from the player
+* score - shows the current score
 
-Count and sum the scores of a bowling game for one player. For this challenge, you do _not_ need to build a web app with a UI, instead, just focus on the logic for bowling (you also don't need a database). Next end-of-unit challenge, you will have the chance to translate the logic to Javascript and build a user interface.
 
-A bowling game consists of 10 frames in which the player tries to knock down the 10 pins. In every frame the player can roll one or two times. The actual number depends on strikes and spares. The score of a frame is the number of knocked down pins plus bonuses for strikes and spares. After every frame the 10 pins are reset.
-
-As usual please start by
-
-* Forking this repo
-
-* Finally submit a pull request before Monday week at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday week at 9am. 
-
-___STRONG HINT, IGNORE AT YOUR PERIL:___ Bowling is a deceptively complex game. Careful thought and thorough diagramming — both before and throughout — will save you literal hours of your life.
-
-## Focus for this challenge
-The focus for this challenge is to write high-quality code.
-
-In order to do this, you may pay particular attention to the following:
-* Using diagramming to plan your approach to the challenge
-* TDD your code
-* Focus on testing behaviour rather than state
-* Commit often, with good commit messages
-* Single Responsibility Principle and encapsulation
-* Clear and readable code
-
-## Bowling — how does it work?
-
-### Strikes
-
-The player has a strike if he knocks down all 10 pins with the first roll in a frame. The frame ends immediately (since there are no pins left for a second roll). The bonus for that frame is the number of pins knocked down by the next two rolls. That would be the next frame, unless the player rolls another strike.
-
-### Spares
-
-The player has a spare if the knocks down all 10 pins with the two rolls of a frame. The bonus for that frame is the number of pins knocked down by the next roll (first roll of next frame).
-
-### 10th frame
-
-If the player rolls a strike or spare in the 10th frame they can roll the additional balls for the bonus. But they can never roll more than 3 balls in the 10th frame. The additional rolls only count for the bonus not for the regular frame count.
-
-    10, 10, 10 in the 10th frame gives 30 points (10 points for the regular first strike and 20 points for the bonus).
-    1, 9, 10 in the 10th frame gives 20 points (10 points for the regular spare and 10 points for the bonus).
-
-### Gutter Game
-
-A Gutter Game is when the player never hits a pin (20 zero scores).
-
-### Perfect Game
-
-A Perfect Game is when the player rolls 12 strikes (10 regular strikes and 2 strikes for the bonus in the 10th frame). The Perfect Game scores 300 points.
-
-In the image below you can find some score examples.
-
-More about ten pin bowling here: http://en.wikipedia.org/wiki/Ten-pin_bowling
-
-![Ten Pin Score Example](images/example_ten_pin_scoring.png)
+## Testing
+-----------------------
+To run rspec tests
+~~~~
+$> rspec
+~~~~
--- a/lib/frame.rb
+++ b/lib/frame.rb
@@ -39,6 +39,10 @@ class Frame
     @frame_type == :spare
   end
 
+  def open_frame?
+    @frame_type == :open
+  end
+
   private
 
   def invalid_roll?(roll)
@@ -82,7 +86,7 @@ class Frame
 
     @frame_type = :strike if pins_downed == TEN_PINS && first_roll?
     @frame_type = :spare if second_roll? && pins_downed + @rolls[FIRST_ROLL] == TEN_PINS
-    @frame_type = :open_frame if second_roll? && pins_downed + @rolls[FIRST_ROLL] != TEN_PINS
+    @frame_type = :open if second_roll? && pins_downed + @rolls[FIRST_ROLL] != TEN_PINS
   end
 
   def second_roll?
--- a/lib/frame.rb
+++ b/lib/frame.rb
@@ -55,10 +55,11 @@ class Frame
     process_first_roll(pins_downed)
   end
 
-  def process_first_roll(pins_downed)
-    return unless first_roll?
 
-    @rolls[FIRST_ROLL] = pins_downed
+  def process_final_roll(pins_downed)
+    return unless last_roll?
+
+    @rolls[LAST_ROLL] = pins_downed
   end
 
   def process_second_roll(pins_downed)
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -1,65 +1,28 @@
...
+~~~~
+git clone git@github.com:nyahehnagi/bowling-challenge-ruby.git
+~~~~
+
+## Testing
+-----------------------
+To run rspec tests
+~~~~
+$> rspec
+~~~~
```
winters145, line 28: Maybe some info or instruction on how to run the program would be helpful

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def invalid_pins?(pins_downed)
+    pins_downed.negative? || pins_downed > TEN_PINS
+  end
+
+  def to_many_pins?(pins_downed)
+    return false if @frame_type == :strike
+
+    pins_downed > pins_left
+  end
```
winters145, line 59: Lots of edge cases covered

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    @rolls << pins_downed
+  end
+
+  def process_frame_type(pins_downed)
+    return unless @frame_type.nil?
+
+    @frame_type = :strike if pins_downed == TEN_PINS && first_roll?
+    @frame_type = :spare if second_roll? && pins_downed + @rolls[FIRST_ROLL] == TEN_PINS
+    @frame_type = :open if second_roll? && pins_downed + @rolls[FIRST_ROLL] != TEN_PINS
```
winters145, line 88: I like the setting of @frame_type

lib/last_frame.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require './lib/frame'
+
+class LastFrame < Frame
```
winters145, line 5: Interesting use of a sub-class, impressive! (Although I'm finding it quite hard to follow exactly what's happening!)

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+class Frame
+  TEN_PINS = 10
+  FIRST_ROLL = 0
+  SECOND_ROLL = 1
```
winters145, line 6: Nice use of constants to keep the code clear and obvious as to what is being referenced

lib/scorecard.rb
```diff
@@ -0,0 +1,81 @@
...
+    frame.strike_frame? ? STRIKE : SPARE
+  end
+
+  def bonus_points(frame_no, rolls_ahead)
+    frames_ahead = @frames.select { |frame_key, _frame| frame_key >= frame_no }
+    
+    frame_scores = frames_ahead.map { |_frame_key, frame| frame.all_rolls }.flatten
+
+    frame_scores = rolls_ahead == SPARE ? frame_scores.drop(2) : frame_scores.drop(1)
+    frame_scores.first(rolls_ahead).count == rolls_ahead ? frame_scores.first(rolls_ahead).sum : ZERO_SCORE
```
winters145, line 79: Clever code, but quite hard to read / understand what is actually being done here

lib/scorecard.rb
```diff
@@ -0,0 +1,81 @@
...
+    if next_frame?
+      @frames[@frames.count + 1] = last_frame_next? ? @last_frame_class.new : @frame_class.new
+    end
+    @frames[@frames.keys.last]
+  end
+
+  def next_frame?
+    return true if @frames.empty?
+    return false if @frames.count == LAST_FRAME
+    return true if @frames[@frames.keys.last].frame_complete?
```
winters145, line 59: Clever bit of iteration over the frames hash

spec/frame_spec.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
```
winters145, line 1: Incredibly wide-ranging tests, great coverage

winters145 (commented):
Great code! Really well extracted methods and classes - SRP followed to a tee! Very in-depth testing too, all possible edge cases seem to have been covered, good job!

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
```
nelsonclaire, line 1: Love the layout of the classes, easy to read and understandable.

README.md
```diff
@@ -1,65 +1,28 @@
...
+----------------------
+## Installation
+----------------------
+Install gems with [bundler](https://bundler.io/ "bundler") 
+~~~~
+$> bundle install 
+~~~~
+To install code from the latest source
+~~~~
+git clone git@github.com:nyahehnagi/bowling-challenge-ruby.git
```
nelsonclaire, line 20: is this correct (thinking it is git@github.com/nyahehnagi/bowling-challenge-ruby.git)?

README.md
```diff
@@ -1,65 +1,28 @@
...
-### Perfect Game
-
-A Perfect Game is when the player rolls 12 strikes (10 regular strikes and 2 strikes for the bonus in the 10th frame). The Perfect Game scores 300 points.
-
-In the image below you can find some score examples.
-
-More about ten pin bowling here: http://en.wikipedia.org/wiki/Ten-pin_bowling
-
-![Ten Pin Score Example](images/example_ten_pin_scoring.png)
+# Bowling Challenge
```
nelsonclaire, line 1: Did you consider doing any modelling before starting the code and capturing it?

lib/last_frame.rb
```diff
@@ -0,0 +1,45 @@
...
+  LAST_ROLL = 2
+  
+  def frame_complete?
+    return @rolls.count == 3 if bonus_frame?
+
+    super
+  end
+
+  def log_roll(pins_downed)
```
nelsonclaire, line 14: Good clear error messages

lib/scorecard.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def strike_or_spare(frame)
+    frame.strike_frame? ? STRIKE : SPARE
+  end
+
+  def bonus_points(frame_no, rolls_ahead)
+    frames_ahead = @frames.select { |frame_key, _frame| frame_key >= frame_no }
+    
+    frame_scores = frames_ahead.map { |_frame_key, frame| frame.all_rolls }.flatten
```
nelsonclaire, line 76: Neat - now need to look up what these do!

spec/features/logging_rolls_to_scorcard_spec.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+describe 'logging rolls to the scorecard' do
+  it 'first frame, first roll, no strike' do
+    scorecard = ScoreCard.new
```
nelsonclaire, line 5: Are these needed in every test?

spec/frame_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    it 'shows a frame is not a spare frame after an open frame' do
+      frame.log_roll(7)
+      frame.log_roll(1)
+      expect(frame.spare_frame?).to be false
+    end
+  end
+
+  context 'input validation' do
+    it 'does not allow rolls less than 0' do
```
nelsonclaire, line 91: Could this actually be input?

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
...
+    process_roll(pins_downed)
+  end
+
+  def sum_frame
+    @rolls.sum
+  end
+
+  def all_rolls
+    @rolls
+  end
```
JohnForster, line 27: This method seems unnecessary to me, as the rolls can be accessed directly through the `@rolls` instance method.

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
...
+    @frame_type == :strike
+  end
+
+  def spare_frame?
+    @frame_type == :spare
+  end
+
+  def open_frame?
+    @frame_type == :open
+  end
```
JohnForster, line 43: Ideally we want to minimise the amount of state stored in a class. Whether the frame is a strike or spare can be calculated directly from other state (the rolls), so it's best not to duplicate it in a `@frame_type` variable.

JohnForster: Instead, consider a strike? method which works out from the rolls whether or not the frame is a strike.

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  private
+
+  def invalid_roll?(roll)
+    roll.negative? || (roll > @rolls.count)
+  end
```
JohnForster, line 49: I don't think this method is being used.

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
...
+    TEN_PINS - @rolls[FIRST_ROLL]
+  end
+
+  def process_roll(pins_downed)
+    process_second_roll(pins_downed)
+    process_first_roll(pins_downed)
+  end
+
+  def process_second_roll(pins_downed)
+    return unless second_roll?
```
JohnForster, line 71: It would be better to move this conditional into the `process_roll` method, so that you're only ever calling `process_second_roll` when you are processing a second roll.

JohnForster: Same goes for the `return unless first_roll?` below

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+class Frame
+  TEN_PINS = 10
```
JohnForster, line 4: Better to name this constant NUMBER_OF_PINS, because TEN_PINS is just as inflexible as using the number 10. If I asked you to adapt your program for a version of bowling that had 15 pins instead of 10, for example, this constant name wouldn't make sense any more.

lib/frame.rb
JohnForster, line 59: I don't see a reason to make this method dependent on the frame_type. The logic should work whether or not the frame type was a strike.

lib/frame.rb
```diff
@@ -0,0 +1,100 @@
...
+    @frame_type = :strike if pins_downed == TEN_PINS && first_roll?
+    @frame_type = :spare if second_roll? && pins_downed + @rolls[FIRST_ROLL] == TEN_PINS
+    @frame_type = :open if second_roll? && pins_downed + @rolls[FIRST_ROLL] != TEN_PINS
+  end
+
+  def second_roll?
+    return false if first_roll?
+
+    @rolls[SECOND_ROLL].nil?
+  end
```
JohnForster, line 95: If you use the .length property of the `@rolls` away, this method and others can become simpler.

lib/last_frame.rb
JohnForster, line 5: Nice idea to use an inherited class for the last frame!

lib/last_frame.rb
```diff
@@ -0,0 +1,45 @@
...
+    raise 'Pins downed must be between 0 and 10' if invalid_pins?(pins_downed)
+    raise 'Frame complete. Cannot roll again' if frame_complete?
+
+    process_frame_type(pins_downed)
+    process_roll(pins_downed)
+  end
```
JohnForster, line 20: As far as I can see this method is the same as in the Frame class, so shouldn't be duplicated.

lib/scorecard.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+require 'frame'
+
+class ScoreCard
+  LAST_FRAME = 10
+  TEN_PINS = 10
+  ZERO_SCORE = 0
```
JohnForster, line 8: This doesn't need to be a constant.

lib/scorecard.rb
```diff
@@ -0,0 +1,81 @@
...
+  SPARE = 1
+  STRIKE = 2
+
+  def initialize(frame_class = Frame, last_frame_class = LastFrame)
```
JohnForster, line 12: Good dependency injection!

lib/scorecard.rb
```diff
@@ -0,0 +1,81 @@
...
+    @last_frame_class = last_frame_class
+    @frames = {}
+  end
+
+  def current_frame_number
+    @frames.count
+  end
+
+  def log_roll(pins_downed)
+    return :game_complete if game_complete?
```
JohnForster, line 23: Trying to enter a roll to a completed game should probably raise an exception.

lib/scorecard.rb
JohnForster: This could be improved by refactoring into clear methods.

spec/features/logging_rolls_to_scorcard_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  it 'having a bad day.. Gutter Game!' do
+    scorecard = ScoreCard.new
+    20.times do
+      scorecard.log_roll(0)
+    end
+
+    expect(scorecard.score).to eq 0
+    expect(scorecard.current_frame_number).to eq 10
```
JohnForster, line 49: Excellent feature tests.

JohnForster (commented):
This is generally an excellent solution Bromley, well done. There are a few areas that could be simplified with a bit of refactoring, and currently the logic for calculating the score of a frame is split between the Frame class and the Scorecard class. Unifying this might simplify the logic. 

RE: Your question about mocking, the reason we mock our tests is to make our code more flexible. Lets say we were using the logic you have written for an online bowling game. If we wanted to add a mode where there were 15 pins insead of 10, how easy would that addition be to make?

As it stands, the tests for your Scorecard class is very much reliant on the frame class we've already written, and we'd need to change a lot of tests to make sure that it would also work with our new expanded frame class. In a perfect world, we would be able to write a new ExpandedFrame class, and be confident that it would work with the existing Scorecard. Mocking the Frame class helps to give us this confidence.

lib/frame.rb
nyahehnagi: Implemented in the javascript version. Thanks for the feedback

lib/frame.rb
nyahehnagi: implemented in the javascript version.  Although not finished I don't think I need to concern myself about 1st and second rolls now as I can now just look at the isStrike/Spare methods

PR #237, opened 2022-02-27
Bowling HF Pull request

Haroon2021:


Strike_and_Spare_methods.rb
```diff
@@ -1,18 +1,17 @@
-def Strike
-    if roll1score == 10
-        print 1
-    else
-        print 0
-    end
+# frozen_string_literal: true
 
+def Strike
```
mmguinness, line 3: To shorten your code, you could look at using ternary operators instead of if/else statements.

testing.rb
```diff
@@ -10,46 +10,62 @@
 
 i = 0
```
mmguinness, line 11: What is 'i' tracking? Maybe give it a more specific name for bowling context.

spec/game_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    it 'pins returns the number of pins knocked down' do
+      game = Game.new
+      expect(game.pins(4)).to eq 4
+    end
+  end
+  describe 'with no strikes or spares' do
+    it 'sums over all the hits' do
+      game = Game.new
+      20.times { game.pins(3) }
+      expect(game.score).to eq 60
```
mmguinness: This is a very good clear test to check a full game works. Next step - it would be good to see score total for each frame, documented on a scorecard.

testing.rb
```diff
@@ -0,0 +1,55 @@
...
+
+
+while (frame <= 10) && (i <= 19)
+
+
+
+  
+  
+  i += 1
+  if i.even?
```
mmguinness, line 29: What is this if statement doing?

testing.rb
```diff
@@ -10,46 +10,62 @@
```
mmguinness, line 10: Is this testing file, a test program to run the whole game? Or part of the final game?

testing.rb
mmguinness: After speaking with you I understand the logic. This approach works well, I had not thought of the problem in this way.

lib/game.rb
```diff
@@ -7,13 +7,27 @@
...
+    @rolls << pins
+    p @rolls
   end
 
-  def pins(n)
-    @score += n
+  def score
+    #@rolls.reduce(:+)  
+    result = 0
+    rollindex = 0
```
mmguinness, line 21: Really interesting approach to use a roll index. Thanks for the link!

spec/game_spec.rb
```diff
@@ -1,21 +1,34 @@
...
+    expect(game.score).to eq 20
+  end
+
+  it 'Can roll a spare' do
+    game = Game.new
+    game.roll(5)
+    game.roll(5)
+    game.roll(3)
+    17.times{game.roll(0)}
+    expect(game.score).to eq 16
```
mmguinness, line 32: Interesting approach to test a full game to check the spare score. Very clear.

PR #240, opened 2022-02-27
Bowling challenge, first attempt - Jack W

winters145:


lib/bowling.rb
```diff
@@ -0,0 +1,39 @@
+class Bowling
```
nyahehnagi, line 1: as discussed, worth looking at the problem by looking at the smallest possible element of the problem. So in this case, a roll or maybe a frame. This would then help later on with the development of the code

lib/bowling.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def initialize
+    @total_score = 0
+  end
+
+  def total_frames(frame_scores)
+    raise 'Minimum of 10 frames must be entered' if frame_scores.count < 10
+    # p frame_scores
+    frame_scores.map.with_index do |frame, index|
+      unless [9, 10, 11].include?(index)
```
nyahehnagi, line 12: Magic Numbers... consider using constants when needing to refer to numbers. This gives the numbers meaning and helps the code read better

lib/bowling.rb
```diff
@@ -0,0 +1,39 @@
...
+  attr_reader :total_score, :frame
+
+  def initialize
+    @total_score = 0
+  end
+
+  def total_frames(frame_scores)
+    raise 'Minimum of 10 frames must be entered' if frame_scores.count < 10
+    # p frame_scores
+    frame_scores.map.with_index do |frame, index|
```
nyahehnagi, line 11: Nice use of map_with_index. This is the beginnings of functional programming. Well worth reading up on it. The better you get with this style of programming the cleaner and more readable your code will become

lib/bowling.rb
```diff
@@ -0,0 +1,39 @@
...
+      unless [9, 10, 11].include?(index)
+        if back_to_back_strike?(frame, frame_scores, index)
+          frame_scores[index] = [frame[0], (frame_scores[index + 1].sum + frame_scores[index + 2].sum)]
+        elsif strike?(frame)
+          frame_scores[index] = [frame[0], frame_scores[index + 1].sum]
+        elsif spare?(frame)
+          frame_scores[index] = [frame[0], (frame[1] + (frame_scores[index + 1][0]))]
+        end
```
nyahehnagi, line 19: Lots of if statements here. Can you think of an alternative way to do this? Perhaps you could call methods and use guard clauses. This would help make the code more readable

lib/bowling.rb
```diff
@@ -0,0 +1,39 @@
...
+      end
+    end
+    # p frame_scores
+    frame_scores.each { |scores| @total_score += scores.sum }
```
nyahehnagi, line 23: you could use map or inject here .. have a look at how you would do that.

README.md
```diff
@@ -1,37 +1,21 @@
 Bowling Challenge in Ruby
 =================
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday week
+For this challenge I followed a TDD approach to write a program that calculates the total score for a game of bowling, of a single player from a given array of pins knocked down by each roll.
```
nyahehnagi, line 4: Good README. Nice to see. A complete README really helps the reviewer work out what's going on. Salut!

spec/bowling_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'bowling'
+
+describe Bowling do
+  let(:game) { described_class.new }
+
+  describe '#total_frames' do
```
nyahehnagi, line 6: As you are using a very specific format for your bowling game, perhaps add additional tests and appropriate code to ensure the input is in the right format.  Your code will fail if the inner arrays are not in the right format

nyahehnagi (commented):
Code works fine. I've tested some other scenarios. Key take away for me is readability. It could be clearer especially since you have used arrays within arrays :-)
Also I don't feel this code could be extended that easily, so I'd suggest pulling out Frames as a separate class/module, this would then allow the code to follow the Single Responsbility Principle. As right now, everything is wrapped up together
Finally, the method of roll input is very bespoke and does not really follow how bowling works, I appreciate it was to help with the calculation. If anything there could be a manager/game class wrapping all this up and ensuring the data sent to the Bowling class is sound

PR #244, opened 2022-02-28
Chris's Bowling Challenge

CTJMID:
Probably could use some class separation...


PLANNINGREADME.md
```diff
@@ -0,0 +1,13 @@
+Plan to build this game, bit by bit, using TDD:
+
```
trushnarabadia, line 2: Great and detailed plan, very clear and to the point

lib/game.rb
```diff
@@ -0,0 +1,93 @@
...
+        @score = 0
+        @frame_counter = 0
+        @spare_counter = 0
+        @extra_roll = 0
+        @results = []
+        @frame = []
+    end
+
+    def roll(pins)
+        fail 'This game is over' if @frame_counter == FRAME_MAX
```
trushnarabadia, line 17: Good idea to let the player know when the game is over rather than just stopping the game.

lib/game.rb
```diff
@@ -0,0 +1,93 @@
...
+        running_score
+    end
+
+    def bonus_ball_spare
+        @frame_counter -= 1
+        @extra_roll = 1
+    end
+    
+
+    def running_score
```
trushnarabadia, line 86: This is a really nice part for the user, very friendly

spec/game_spec.rb
```diff
@@ -0,0 +1,116 @@
+require './lib/game.rb'
+
```
trushnarabadia, line 2: very extensive testing, covered all the scenarios required in order for the game to be played.

lib/game.rb
```diff
@@ -0,0 +1,93 @@
+class Game
+
```
trushnarabadia, line 2: Code is very well refactored, code is very DRY. Looks great!

spec/game_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+            expect(subject.roll(10)).to eq "The game is finished - you scored 48 points"
+        end
+    end
+
+    describe 'a perfect game' do
+      it 'returns a finishing message and 300 when a perfect game is bowled' do
+        11.times{subject.roll(10)}
+
+        expect(subject.roll(10)).to eq "The game is finished - you scored 300 points"
+        end
```
trushnarabadia, line 114: I have run RSpec and all tests have passed.

lib/game.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    def bonus_ball_spare
+        @frame_counter -= 1
+        @extra_roll = 1
+    end
+    
+
+    def running_score
+      return "The game is finished - you scored #{@score} points" if @frame_counter == FRAME_MAX
```
leoht, line 87: Here rather than returning human-readable messages, the return value of this method (and therefore, the result returned by `roll`) could maybe be the new score? Or if you want to also return the information about a potential bonus, you could return a Hash (or maybe even a custom result class):
```irb
return { score: @score, bonus: :strike } # for example
```

This has the advantage of being more "structured" than a plain string (in which we could have typos, etc), so the caller of this class will be able to manipulate and read the result and use it if needed - also, it avoids mixing concerns into the `Game` class (human-readable messages and deciding what to print depending on the result would be more part of presentation/display, so it would probably belong in a different class)

spec/game_spec.rb
```diff
@@ -0,0 +1,116 @@
+require './lib/game.rb'
+
+describe Game  do
+
+    describe 'score' do
+        it 'adds the scores of different rolls, and returns a score' do
+        subject.roll(1)
+        subject.roll(2)
+        expect(subject.score).to eq 3
```
leoht, line 9: Rather than accessing the value of the internal attribute `score` (exposed by an attr_reader/accessor), prefer to assert on the return value of `roll`, since this is the behaviour of the method we're looking to test

spec/game_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+        end
+    end
+
+    describe 'end game'  do
+        it 'ends the game after max rolls, and returns the final score' do
+            ((Game::FRAME_MAX * 2) - 1).times{subject.roll(1)}
```
leoht, line 15: This is slightly harder to read - maybe you could rewrite it such as:
```ruby
max_roll_count = ((Game::FRAME_MAX * 2) - 1)
max_roll_count.times { subject.roll(1) }
```

spec/game_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+            expect(subject.frame_counter).to eq 1
+        end
+    end
+
+    describe 'strike identifier' do
+        it 'changes the frame number as soon as a 10 is rolled' do
+            subject.roll(1)
+            subject.roll(1)
+            subject.roll(10)
+            expect(subject.frame_counter).to eq 2
```
leoht, line 41: Same remark as above here - prefer testing the return value of `roll` (or verify if it raises an error, if needed, etc) rather than accessing the internal value `.frame_counter` to verify your test

spec/game_spec.rb
leoht: Also it looks like this value is used at different places in the tests, maybe you could refactor it into a `let` block at the beginning of the test suite, so it's available everywhere?

spec/game_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+        it 'give a two bonus roll on final frame if a strike' do
+            ((Game::FRAME_MAX * 2) - 2).times{subject.roll(1)}
+            subject.roll(10)
+            subject.roll(10)
+            expect(subject.roll(10)).to eq "The game is finished - you scored 48 points"
+        end
+    end
+
+    describe 'a perfect game' do
+      it 'returns a finishing message and 300 when a perfect game is bowled' do
```
leoht, line 110: It seems the indentation is slightly off compared to the other tests? You can use rubocop to check and automatically fix this kind of format/style issues

lib/game.rb
```diff
@@ -0,0 +1,93 @@
...
+        @extra_roll = 0
+        @results = []
+        @frame = []
+    end
+
+    def roll(pins)
+        fail 'This game is over' if @frame_counter == FRAME_MAX
+        @results << pins
+        @frame  << pins
+        bonus_determiner
```
leoht, line 20: Looks like so far you have one class `Game` which calculates everything - maybe you could start to think on how to extract some concerns to other class, and have `Game` delegate some work to these new classes. For example, everything related to bonus calculation could go into its own class, the code of `roll` could look like this:

```irb
def roll
   fail 'This game is over' if @frame_counter == FRAME_MAX
  @results << pins
  @frame  << pins
  
  @score += @bonus_calculator.calculate_added_bonus(pins)

  # etc
end
```

And following the same process, you can extract maybe the part that calculates the regular frame score, etc. Note that this is one way of splitting this class, but you could find other ways, and this is fine too!

leoht (commented):
Nice work! I left a few comments on how to refactor the code, it'd probably be good to split the `Game` class into a few smaller ones - let me know if you have any questions on the feedback :)

PR #247, opened 2022-02-28
bowling-challenge-ruby

Jaroslaw-Roman-Cudzich:
"Created tests, variables, a class, and methods for rolling a gutter game, all 1s, a spare, and a strike."


lib/bowling_game.rb
```diff
@@ -0,0 +1,29 @@
...
+        10.times do
+          if @rolls[rollIndex] + @rolls[rollIndex + 1] == 10
+            result += @rolls[rollIndex] + @rolls[rollIndex + 1] = @rolls[rollIndex + 2]
+          else
+            result += @rolls[rollIndex] + @rolls[rollIndex + 1]
+          end
+          rollIndex += 2
+        end
+        result
+    end
```
SarahM55, line 23: Would love to know more about how you got this to work - I tried to use the same method but no matter what I did, I couldn't get the 'rollIndex' part to work! This is a really nice :)

spec/bowling_game_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        expect(@game.score).to eq 16
+    end
+
+    it "is able to roll a strike" do
+         game = BowlingGame.new
+         @game.roll 10
+         @game.roll 4
+         @game.roll 3
+         16.times(@game.roll 0)
+         expect(@game.score).to eq 24
```
SarahM55, line 35: This looks really good, nice succinct way of writing the tests. To avoid repetition of 'game = Bowling.new', you could have done:

'before do 
  @game = Bowling.new
end'

at the top of your code, that way you could have just called '@game' in each test instead of calling 'game = Bowling.new' each time :D

lib/bowling_game.rb
SarahM55: If you'd had more time, you could have refactored this a bit by breaking it up into new methods like 'spare?', 'strike?' etc - looks like you started doing that below!

SarahM55 (commented):
This is a really good start! Looks like you've clearly understood the logic behind the rules of bowling and have taken your time in implementing the tests :)

PR #250, opened 2022-02-28
Bowling_challenge

Rakshak81:


spec/game_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+describe Game do
+#   describe '#score' do
+#     it 'calculates the score' do
+#       game = Game.new
+#       20.times {game.hits(2)}
+#       expect(game.score).to eq 40
+#     end
+#   end  
+# end
+it 'can do' do
```
nelsonclaire, line 11: It would be good to add a bit more detail on what the test is for so it is easy to understand and separate it from other tests

spec/game_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  game = Game.new
```
nelsonclaire, line 12: Clear test for gutter game

lib/game.rb
```diff
@@ -0,0 +1,35 @@
+class Game
+
+  # attr_reader :score
+
+  def initialize
```
nelsonclaire, line 5: Good use of an array to hold the number of pins knocked down in a roll

spec/game_spec.rb
```diff
@@ -14,4 +14,9 @@
   expect(game.score).to eq 0
 end
 
+it 'can do' do
```
nelsonclaire, line 17: As before would be good to show that this is testing that it can add up correctly for a game with no strikes or spares

PR #252, opened 2022-03-27
Nico's bowling scorecard calc

NicolaCortese:


lib/scorecard.rb
```diff
@@ -0,0 +1,70 @@
+class Scorecard
+
+  def initialize (scores)
+    @scores = scores
+    @total_score = 0
+    @spare = false
+    @strike = false
+    @strike2 = false
```
PKilgarriff: I really like this way of doing it - makes it clear upfront what you are aiming for.

lib/scorecard.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def score
+    @scores.each_with_index do |(frame, roll), index|
+      if index == 9
```
PKilgarriff: At the moment this is a magic number - could it be a constant like FINAL_FRAME_INDEX

PKilgarriff (commented):
As discussed, think the readability would benefit from some encapsulation (but not too far like my approach). You've sensibly approached this by doing the most straightforward thing first (input fully working scorecard), which gives a great base to work from and add functionality. Looking forward to pair work later!

lib/final_frame.rb
```diff
@@ -0,0 +1,61 @@
...
+
+# this class is instatiated by the Game class
+# whenever it reaches the 10th frame
+class FinalFrame
+  attr_reader :strike, :spare, :rolls
+
+  def initialize
+    @rolls = []
+    @strike = false
+    @spare = false
```
alicelieutier: These variables are derived state, meaning their value can be derived from other instance variables in this class.

Storing derived state is like storing duplicate information, and it can lead to bugs (for example if one is updated and not the other).

A better approach is then to have methods that calculate this information on the fly based on the state. so something like:
```rb
def strike?
  @rolls[0] == 10
end
```
etc.

alicelieutier: You've done this with other methods in this class, so this is just taking it a bit further ;)

lib/final_frame.rb
```diff
@@ -0,0 +1,61 @@
...
+    if !@spare && !@strike && complete?
+      true
+    elsif @spare && !@bonus.empty?
+      true
+    elsif @strike && @bonus.count == 2
+      true
+    else
+      false
+    end
+  end
```
alicelieutier: Should some of these methods be private?

If the answer is: "No, because all of them are necessary for another class to use", then I wonder if the class is too tightly coupled with another class.

lib/frame.rb
```diff
@@ -0,0 +1,59 @@
...
+# this class is instatiated by the Game class
+# on instatiation, and subsequently on every roll
+# when the prior frame is complete
+class Frame
+  attr_reader :strike, :spare, :rolls
+
+  def initialize
+    @rolls = []
+    @strike = false
+    @spare = false
```
alicelieutier: Same comment as above.
Also, should the tenthFrame inherit for the frame class so you don't have to rewrite some of the code?

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+
+# this class initiates the whole game,
+# it is dependant on both the frame and final frame
+class Game
+  attr_reader :scorecard
+
+  def initialize(frame = Frame.new)
+    @current_frame = frame
+    @frame_count = 1
+    @scorecard = {}
```
alicelieutier: I'm wondering why this is a hash and not an array 🤔

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def check_frame
+    return unless @current_frame.complete? && @frame_count != 10
+
+    @frame_count += 1
+    @current_frame = if @frame_count == 10
+                       FinalFrame.new
+                     else
+                       Frame.new
+                     end
```
alicelieutier: This type of if/else is often expressed as a ternary expression:
```rb
@current_frame = @frame_count == 10 ? FinalFrame.new : Frame.new
```

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def score
+    total_score = 0
+
+    @scorecard.each do |_frame_count, frame|
+      total_score += frame.frame_score
+    end
+
+    total_score
+  end
```
alicelieutier: Should this be private?

alicelieutier: Nice way of counting the score by the way!

If you wanted to make it even more "functional style", you could write:
```rb
@scorecard.map {|_, frame| frame.frame_score }.sum
```

alicelieutier: using `_` here to show we are not using that variable.

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+require_relative './final_frame'
+
+# this class initiates the whole game,
+# it is dependant on both the frame and final frame
+class Game
+  attr_reader :scorecard
+
+  def initialize(frame = Frame.new)
+    @current_frame = frame
+    @frame_count = 1
```
alicelieutier: Do you need the frame count? Could you just check how many frames are already in the hash and add 1? (or make it an array...)

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+    @scorecard[@frame_count] = @current_frame
+
+    # add pins to the current playing frame
+    @current_frame.add_roll(pins)
+
+    # check if the previous frame needs bonus
+    @scorecard[@frame_count - 1]&.add_bonus(pins)
+
+    # check if the two prior frame needs bonus
+    @scorecard[@frame_count - 2]&.add_bonus(pins)
```
alicelieutier: I like this. Nice delegation to the Frame class!

lib/final_frame.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def add_bonus(pins)
+    # if spare and we haven't added a bonus yet
+    @bonus << pins if @spare && @bonus.empty?
+    # if strike
+    @bonus << pins if @strike && @bonus.count < 2
+  end
+
+  def frame_score
```
alicelieutier: Detail, but given the class is already named Frame, surely this method could just be named "score" ?

lib/game.rb
alicelieutier: You can extract this to a `distribute_bonus_rolls(pins)` method.

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    # after adding a roll we check if the frame
+    # is complete and we start a new one
+    check_frame
+    print_score
```
alicelieutier: I'm not entirely sure why these methods are extracted vs the rest of the code above. It feels a bit like you thought - I have too many lines already, I'll put the rest in methods, rather than trying to extract meaningful parts.
OR - maybe the naming could be more meaningful, like "create_new_frame_if_needed" and "print_score_if_game_is_complete"

Otherwise, you could do something like:
```rb
if new_frame_needed
  @frame_count += 1
  @current_frame = @frame_count == 10 ? FinalFrame.new : Frame.new
 end
 
 if complete?
   print_score
 end
```

Where `new_frame_needed` and `print_score` are adjusted.

The reason for this is keeping the main flow of control in this method.

alicelieutier: Another reason for my proposing this is that it means that only this method changes instance variables, while all other helper method do not change them. This way by looking at just this method I can see all the mutations, without anything being "hidden".

This is a difficult concept to explain, so sorry if I'm not doing great :/

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+                       Frame.new
+                     end
+  end
+
+  def print_score
+    # if game is complete
+    if @frame_count == 10 && @current_frame.score_complete?
+      p "Game finished! Your total score is #{score}"
+    else # otherwise print the score for the complete frames
+      p "Your current score is #{score}"
```
alicelieutier: `p` is great for debugging, but when displaying to the user you should use `puts`.

alicelieutier (commented):
Really good separation of concern, and great tests. 
Overall, your code is easy enough to read, too.
You can go further with refactoring (this is bowling, after all, the never ending refactoring challenge), I added some pointers in the code.

Let me know if anything doesn't make sense!

PR #264, opened 2022-03-28
Adam's Bowling Challenge

adamwoodcock98:
- [x] Add a spare bonus
- [x] The players moves onto the next frame
- [x] Total score is increased
- [x] Score is increased
- [x] The player rolls again
- [x] The pins are input and added to a score
- [x] Player rolls
- [x] Add a strike bonus
- [ ] Add an extra turn on the tenth frame
- [ ] Incorporate a perfect game


lib/frame.rb
```diff
@@ -0,0 +1,36 @@
...
+  def rolled(pins)
+    @rolls << pins
+  end
+
+  def score
+    @score = @rolls.sum
+  end
+
+  def next_turn?
+    @rolls.length == 2 || strike?
```
joe-thornton, line 20: Very neat logic

lib/scorecard.rb
```diff
@@ -0,0 +1,53 @@
...
+      calculate_total
+    end
+  end
+
+  def turn
+    @current_turn
+  end
+
+  def calculate_total
+    @total_score = 0
```
joe-thornton, line 29: I suspect there may be issues setting to zero here and in the initialize method - I guess whether it should be set to zero here depends on whether this method is a running total or the total just by each frame

lib/scorecard.rb
```diff
@@ -0,0 +1,53 @@
...
+    calcualate_bonus
+    @frames.each { |frame| @total_score += (frame.score + frame.bonus_score) }
```
joe-thornton, line 31: This method returns an array of frames currently - should it return 'total_score'?

lib/scorecard.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  private
+
+  def calcualate_bonus
```
joe-thornton, line 36: From irb, currently bonus is being calculated as 10 if someone rolls a strike and then 2 subsequent strikes

spec/units/scorecard_spec.rb
```diff
@@ -0,0 +1,84 @@
+require 'scorecard.rb'
+
+describe Scorecard do
```
joe-thornton, line 3: It might be helpful to increase the test coverage to include full games (e.g. perfect game), especially where the final frame has 3 rolls

joe-thornton (commented):
Great effort. Nice separation of responsibilities between classes. Some very neat, concise logic in the methods.

Suggest increasing the test coverage for entire games.

PR #267, opened 2022-04-23
Bowling Challenge in Ruby

Josenewmano:
Full functionality, and fairly concise, but a little bit naughty in terms of process - completed on paper on the bus before being typed up

### DIFF
--- a/score_calculation_interface.rb
+++ b/score_calculation_interface.rb
@@ -1,3 +1,4 @@
+## Inputting the scores for the first nine rounds
 frame_count = 0
 scores = []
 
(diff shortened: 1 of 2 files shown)


score_calculation_interface.rb
```diff
@@ -0,0 +1,55 @@
...
+    bonus += scores[bonus_count + 1][0]
+  end
+
+  bonus_count += 1
+end
+
+total = scores.flatten.sum + bonus
+
+print "\n\nYour total score for this game was #{total}!"
+puts ' A perfect game! Wow!' if total == 300
```
david-ewan-campbell, line 59: Very enjoyable to use, user-friendly 'classic' Ruby interface that actually had the functionality of telling you your score frame by frame and told you which frame you were on (unlike my code) ...also having the functionality of strikes so if you rolled ten you moved to next frame. Although it wasn't following a TDD process, I think it's great you could write this functioning scorecard that tells you each round and also added a bonus score in the final round. Like you Joe i'd like to add a running total to my code.

david-ewan-campbell (commented):
Very nice functioning scorecard program Joe!

PR #269, opened 2022-04-24
Joe McNally bowling challenge completed

jmcnally17:
Joe McNally

# User stories 

- [x] User story 1: "I want to calculate the score of a gutter game"
- [x] User story 2: "I want to calculate the score of a game with no spares or strikes"
- [x] User story 3: "I want to calculate the score of a game with a spare"
- [x] User story 4: "I want to calculate the score of a game with a strike"
- [x] User story 5: "I want to calculate the score of a game with a bonus frame"
- [x] User story 6: "I want to calculate the score of a game with consecutive strikes"
- [x] User story 7: "I want to calculate the score of a perfect game"

# README checklist

- [x] how to install,
- [x] how to run,
- [x] and how to test your code?


Gemfile
```diff
@@ -0,0 +1,13 @@
+source 'https://rubygems.org'
+
+ruby '3.1.1'
```
almorcrette, line 3: Thought: Is this version of ruby really necessary? It isn't. would it then be better not to require it?

I understand that Gemfile is there to help another user have the same background software as the developer to ensure that the program also works for them. You used 3.1.1. so you are telling other users to do the same.

On the other hand, installing the latest ruby takes time (in fact for some reason, it's taking forever for me - I am running 3.0.0 and have 3.1.0 installed too but not 3.1.1), so if it's not required, would it be better to allow users to run your program on an earlier version of ruby?

I'm not sure how you would decide which earlier version you would use.

lib/game.rb
```diff
@@ -0,0 +1,38 @@
...
+  private
+
+  def frame_score(frame, index)
+    return 0 if index > 9                                    # don't count the bonus on their own
+    frame_score = frame.inject(0,:+)
+    frame_score += strike_or_spare(frame, index) if frame_score == 10 # if a strike or spare
+    frame_score
+  end
+
+  def strike_or_spare(frame, index)
```
almorcrette, line 23: Could be clearer with naming of this. e.g. bonus_scoring

lib/game.rb
```diff
@@ -0,0 +1,38 @@
...
+    if frame.length == 1                                    # if a strike
+      strike_score(index)
+    else                                                    # if a spare
```
almorcrette, line 26: You could potentially protect this method being used in future versions of the code by including a guard against frames scoring less than 10, where they return 0

lib/game.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+    score
+  end
+
+  private
+
+  def frame_score(frame, index)
+    return 0 if index > 9                                    # don't count the bonus on their own
+    frame_score = frame.inject(0,:+)
+    frame_score += strike_or_spare(frame, index) if frame_score == 10 # if a strike or spare
```
almorcrette, line 19: Could integrate the if statement into the strike_or_space method. Benefit of this is that strike_or_spare could be applied to all frames without generating erroneous bonuses. See my comment below about a guard for non-10-scoring frames in the strike_or_space method

almorcrette (commented):
- [x] 100% test coverage
- [x] Fully linted (with rubocop)

- [x] Using diagramming to plan approach to the challenge - *recommended including it as a picture in your README*
- [x] TDD
- __not sure__ Focus on testing behaviour rather than state
- [x] Committing often, with good commit messages
- [x] Single Responsibility Principle and encapsulation
- [x] Clear and readable code *see note about including guard clause in strike_or_space and renaming it*

Also see note about requiring versions of Ruby. Not sure the answer to this.

PR #270, opened 2022-04-24
Jordan Manu Bowling Challenge

JordanManu:
## Features:

* [X] Strikes
* [X] Spares
* [X] 10th Frame
* [X] Gutter Game
* [X] Perfect Game

## Readme Checklist

* [X] how to install
* [X] how to run
* [X] how to test code

### DIFF
--- a/lib/scorecard.rb
+++ b/lib/scorecard.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+class Scorecard
+  attr_reader :frames, :next_frame
+
+  def initialize(frames)
+    @frames = frames
+  end
+
+  def final_score
+    final_score = 0
+    @frames.each_with_index do |frame, index|
+      final_score += frame.sum
+      final_score += strike_bonuses(frame, index)
+      final_score += spare_bonuses(frame, index)
+    end
+    final_score
+  end
+
+  def strike_bonuses(frame, index)
+    return 0 unless strike?(frame)
+    return 0 if index == 9
+
+    @next_frame = @frames[index + 1]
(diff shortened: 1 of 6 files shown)


lib/scorecard.rb
```diff
@@ -0,0 +1,50 @@
...
+    if strike?(next_frame)
+      @next_frame[0] + @final_frame[0]
+    else
+      @next_frame.sum
+    end
+  end
+
+  def strike?(frame)
+    frame.include?(10)
+  end
```
mcsuGH, line 37: can private methods such as these since they are used for the code but aren't being tested

lib/scorecard.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class Scorecard
+  attr_reader :frames, :next_frame
+
+  def initialize(frames)
```
mcsuGH, line 6: maybe need a way to ensure people add only 10 frames (can add more than 10 currently)

lib/scorecard.rb
```diff
@@ -0,0 +1,50 @@
...
+    @frames = frames
+  end
+
+  def final_score
+    final_score = 0
+    @frames.each_with_index do |frame, index|
+      final_score += frame.sum
+      final_score += strike_bonuses(frame, index)
+      final_score += spare_bonuses(frame, index)
```
mcsuGH, line 15: like the way index is used here!

mcsuGH (commented):
Very good overall, readable & concise, maybe needs more testing or can use private methods

PR #272, opened 2022-04-24
Conor Hope's Bowling challenge

Conor-Developer:


spec/frame_spec.rb
```diff
@@ -0,0 +1,21 @@
+require_relative '../lib/frame'
+
+describe Frame do
+
+  let(:frame) { described_class.new }
+
+  context '#knocked_down_pins' do
+    it 'logs the number of pins knowed down' do
```
monenavarro, line 8: typo here

lib/frame.rb
```diff
@@ -0,0 +1,24 @@
+require_relative './score'
```
monenavarro, line 1: Code looking really good so far! It's DRY and readable, easy to understand. Great job!

PR #280, opened 2022-05-22
bowling-challenge-ruby-Erlantz

ErlantzR:


spec/test_games.rb
```diff
@@ -0,0 +1,43 @@
+def test_game_one
```
jimmy-lyons, line 1: Have you covered enough games? I think there's more permutations in the final frame that could be tested. It's hard to completely assert that everything works as intended from looking at the unit tests.

PR #283, opened 2022-05-23
Stevie Spiegl bowling challenge

S-Spiegl:


spec/frame_spec.rb
```diff
@@ -0,0 +1,46 @@
+require_relative '../lib/frame'
+
+describe Frame do
+    it 'allows a player to input each of their rolls' do
+        subject.input_roll(9)
```
SamButton12, line 5: I think you could take out this test in your refactor step as it is covered by implication in future tests

PR #286, opened 2022-05-23
Tom-bowling-challenge

tomoneill32:


lib/game.rb
```diff
@@ -0,0 +1,111 @@
...
+  def initialize
+    @frame_number = 1
+    @all_rolls = []
+    @strikes = []
+    @spares = []
+    @total_score = 0
+    @score_by_frame = []
+  end
+
+  def run
```
lukestorey95, line 18: Nicely encapsulated code!

lib/eleventhframe.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+# similar to frame, but for the eleventh frame
```
lukestorey95, line 3: As you say, similar to frame so perhaps frames 11 & 12 could have inherited from the Frame class which would have kept these additional classes a little dryer.

lib/game.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+require 'frame'
+require 'eleventhframe'
```
lukestorey95, line 4: Why is eleventh frame required but not twelfth?

lib/eleventhframe.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def initialize(frame_10_spare)
+    @frame_10_spare = frame_10_spare
+    @roll_one = nil
+    @roll_two = nil
+    @strike = false
+    @spare = false
+  end
+
+  def run
```
lukestorey95, line 16: The users rolls in frames 11 and 12 are being printed twice. Believe the score is still being calculated correctly so not a major bug but might confuse the user

lib/frame.rb
```diff
@@ -0,0 +1,59 @@
...
+    @strike
+  end
+
+  def spare?
+    @spare
+  end
+
+  def total
+    if @roll_two.nil?
+      @roll_one
```
lukestorey95, line 38: You have an attribute reader so you don't need the @ as you're able to refer to the attr_reader method rather than the instance variable directly

spec/frame_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+describe Frame do
+  let(:frame) { Frame.new(1) }
+
+  describe '#run' do
+
+    it 'correctly records a strike' do
+      $stdin = UserInput.input([10])
+      frame.run
+      expect(frame.strike?).to eq(true)
```
lukestorey95, line 11: I believe that for regular RSpec best convention is to try and keep tests to one expect statement. Consider the have_attributes matcher to eliminate multiple expectations in one test, but still test the attributes are set correctly.

https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/have-attributes-matcher

lib/frame.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def calculate_strike
+    return unless @roll_one == 10
+
+    puts 'Strike!'
+    @strike = true
+  end
+
+  def calculate_spare
+    return unless total == 10 && @roll_one != 10
```
lukestorey95, line 54: Magic numbers...

lukestorey95 (commented):
Some really great code here, and love to see lots of encapsulation especially in a complex program when you don't want to expose the internal workings. Left a couple of comments that might be useful to think about and one potential bug. Definitely think that with some further refactoring this could be really clear.

PR #288, opened 2022-05-23
Jimmy Bowling Game

jimmy-lyons:
Half finished


lib/round.rb
```diff
@@ -0,0 +1,52 @@
...
+      @basic_score += frame.sum
+    end
+    @basic_score -= final_frame_bonus
+  end
+  
+  def calculate_bonus_score
+    score_frames(final_frame_length: @frames.last.length)
+  end
+
+  def score_frames(final_frame_length:)
```
ErlantzR, line 30: As we discussed, this will need something to work for consecutive strikes

lib/round.rb
```diff
@@ -0,0 +1,52 @@
+class Round
```
ErlantzR, line 1: I wish I would have thought on getting all the frames in the class at once, instead of going one by one. Well thought as that make it more simple

spec/units/round_spec.rb
```diff
@@ -0,0 +1,50 @@
+require './lib/round'
```
ErlantzR, line 1: Good amount of tests to cover all the possible outputs

PR #291, opened 2022-06-19
Creating frame

Hafizalimi7:


spec/game_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'game'
+
+
+RSpec.describe Bowling, "#score" do
```
robimaliqi, line 4: Clear and concise tests

lib/frame.rb
```diff
@@ -0,0 +1,28 @@
+class Frame
+
+  attr_reader :current_frame
+
```
robimaliqi, line 4: Good use of classes and methods, very easy to read and understand

robimaliqi (commented):
Pretty straight forward, well written code, does the job. Not much more to say.

PR #293, opened 2022-06-19
Abhi N Bowling Challenge

abacus2020:


lib/bowling.rb
```diff
@@ -0,0 +1,37 @@
+class Bowling
+  attr_accessor :frames
+
+  def initialize(scorecard)
+    self.frames = scorecard
+  end
```
kaykakaraka, line 6: This use of self is cool

lib/bowling.rb
```diff
@@ -0,0 +1,37 @@
...
+  def total
+    score = 0
+    frames.each_with_index do |frame, index|
+      score += sum_frame(frame)
+      score += sum_frame(frames[index + 1]) if strike?(frame) && index < 9
+      score += first_roll(frames[index + 1]) if spare?(frame) && index < 9
+    end
+    if score == 210
+      score += 90
+    else
```
kaykakaraka, line 33: Could 31 & 32 be in the same line?

spec/bowling_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  it 'should return the correct score, given a spare in the 10th frame' do
+    scorecard = Bowling.new([[1, 2], [3, 4], [5, 1], [2, 3], [4, 5], [1, 2], [3, 4], [5, 1], [2, 3], [5, 5], [10]])
+    result = scorecard.total
+    expect(result).to eq 71
+  end
+  it 'should return a score of 300 given a perfect game' do
+    scorecard = Bowling.new([[10, 0], [10, 0], [10, 0], [10, 0], [10, 0], [10, 0], [10, 0], [10, 0], [10, 0], [10, 0], [10, 10]])
+    result = scorecard.total
+    expect(result).to eq 300
+  end
```
kaykakaraka, line 63: great tests! all methods are tested

lib/bowling.rb
```diff
@@ -0,0 +1,37 @@
+class Bowling
+  attr_accessor :frames
+
```
kaykakaraka, line 3: I think a more developed version of this might allow the user to simply enter their score after each roll, and then the program organises this into the array of arrays

lib/bowling.rb
```diff
@@ -0,0 +1,37 @@
...
+    sum_frame(frame) == 10 unless frame.include? 10
+  end
+
+  def total
+    score = 0
+    frames.each_with_index do |frame, index|
+      score += sum_frame(frame)
+      score += sum_frame(frames[index + 1]) if strike?(frame) && index < 9
+      score += first_roll(frames[index + 1]) if spare?(frame) && index < 9
+    end
```
kaykakaraka, line 30: this is nice and clear

README.md
```diff
@@ -63,3 +63,50 @@ In the image below you can find some score examples.
...
+
+Input: An array of ten frames, each of which is an array of two rolls
+Output: The sum total of points during that game
+
+Input: A score of ten, i.e. a ‘strike’ has occured during a roll
+Output: The points from the next two rolls are added to the total score
+
+Input: A score of ten has occured during a frame, i.e. a spare
+Output: The points from the next roll is added to the total score
+
```
kaykakaraka, line 79: These are good and useful. Would be cool to also have a design image if you had one

kaykakaraka (commented):
Looks good and like you have broken the problem down into something clear and simple. Easy to read.  In any refactoring you could use private methods for anything only called by the total method and think about using self.total to initialise the scorecard.

PR #296, opened 2022-06-20
Iteration 1 of bowling challenge -- still needs refactoring

ak2022dev:
I test-drove from start to finish and think it works for all the scenarios but I do still need to refactor it.


ak2022dev:
Worked further on refactoring

PR #300, opened 2022-06-20
made and

IbrahimPala1:
made and passed a test file for adding points up


spec/counter_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'counter'
+
+RSpec.describe Counter do
+    it "initially reports a zero points" do
```
raphaella-rose, line 4: I like the test for the initial value before testing for total points accumulated because it isolates the functionality. :)

spec/counter_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'counter'
+
+RSpec.describe Counter do
+    it "initially reports a zero points" do 
+        counter = Counter.new 
+        expect(counter.report_points).to eq "You have scored 0 points so far"
+    end
+
+    it "adds points to zero" do
```
raphaella-rose, line 9: Maybe this test could have a clearer title.

raphaella-rose (commented):
I think this is a good start and your tests for the counter class are very gradual which is great for isolating what each method should be doing. :)

PR #308, opened 2022-08-22
Bowling Scoreboard app

ZacMossHK:


lib/frame.rb
```diff
@@ -0,0 +1,53 @@
...
+        return value
+      end
+      @io.puts "Please enter a value of #{max_value} or less"
+    end
+  end
+
+  def add_bonus_points(pinfall)
+    # pinfall is an integer
+    # adds bonus points to rolls that have spares or strikes
+    @bonus_rolls -= 1
```
SophLewDev, line 37: I just need explaining but I'm not sure why it's minus 1.

lib/frame.rb
```diff
@@ -0,0 +1,53 @@
...
+  attr_accessor :rolls, :bonus_rolls, :bonus_points
+
+  def initialize(io = Kernel)
+    @io = io
+    @rolls = []
+    @bonus_rolls = 0
+    @bonus_points = 0
+  end
+
+  def format_roll
```
SophLewDev, line 11: Bit unsure with this as in the Readme there's examples of like '3 3' and then the use of '/' with '5 / '? Again this is probably just something that needs to be explained to me

SophLewDev (commented):
Really good code as always! Just one or two things I probably need clarity with to understand

PR #311, opened 2022-08-22
Bowling-Challenge

SophLewDev:


lib/frame.rb
```diff
@@ -0,0 +1,20 @@
+class Frame
+  attr_reader :first_roll, :second_roll, :strike, :spare
```
stephenfletchtek, line 2: Maybe just access these four values by calling the methods? This could further simplify the class

lib/user_interface.rb
```diff
@@ -0,0 +1,85 @@
...
+  def print_menu
+    @io.puts "Please choose an option"
+    @io.puts "1. Play next frame"
+    @io.puts "2. Show my score"
+    @io.puts "3. End game"
+  end
+
+  def process(option)
+    case option
+      when "1"
```
ZacMossHK, line 29: can be refactored as
```
when "1" then rolls
```

lib/user_interface.rb
```diff
@@ -0,0 +1,85 @@
...
+        score
+      when "3"
+        @io.puts "The game has finished"
+        exit
+      else
+        @io.puts "Please choose one of the listed options"
+    end
+  end
+
+  def rolls
```
ZacMossHK, line 41: this could be refactored into smaller methods. lots of repeated code

spec/frame_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      expect(frame.strike).to eq false
+    end
+    it "returns false when user does not get a spare" do
+      frame = Frame.new(6,3)
+      expect(frame.spare).to eq false
+    end
+  end
+  context "is a strike or spare" do
+    it "returns true when user gets a strike" do
+      frame = Frame.new(10,0)
```
stephenfletchtek, line 16: A strike frame would have 10 and no zero perhaps?

lib/user_interface.rb
```diff
@@ -0,0 +1,85 @@
+require_relative './game'
+
+class UserInterface
+  def initialize(io)
```
ZacMossHK, line 4: this could be ```def initialize(io = Kernel)``` so that you don't have to call Kernel when you want to run this on line 85

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+    spare, strike, consecutive_strike = false
+
+    @frames.each { |frame|
+      score += frame.first_roll if spare
+      score += (frame.first_roll + frame.second_roll) if strike
+      score += (frame.first_roll) if consecutive_strike
+      score += frame.third_roll if frame.is_a?(TenthFrame)
+      score += (frame.first_roll + frame.second_roll)
+      consecutive_strike = strike && frame.strike
+      spare, strike = frame.spare, frame.strike
```
stephenfletchtek, line 22: Could perhaps just call frame.is_spare or frame.is_strike

lib/user_interface.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def print_menu
+    @io.puts "Please choose an option"
+    @io.puts "1. Play next frame"
+    @io.puts "2. Show my score"
+    @io.puts "3. End game"
+  end
+
+  def process(option)
```
ZacMossHK, line 27: not clear from method name what this does

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+    }
+    return score
+  end
+
+
+  def add_frame(first_roll, second_roll = 0, third_roll = 0)
+    return "The game has finished" if @frames.length == 10
+    return @frames << Frame.new(first_roll, second_roll) if @frames.length < 9
```
ZacMossHK, line 30: could this be the last line so returns without return being called? eg.
```
return "The game has finished" if @frames.length == 10
return @frames << TenthFrame.new(first_roll, second_roll, third_roll) if @frames.length == 9
@frames << Frame.new(first_roll, second_roll)
```

lib/tenth_frame.rb
```diff
@@ -0,0 +1,10 @@
+require_relative './frame'
```
stephenfletchtek, line 1: I wonder if you need this as an object? It could be simplified by using a hash for the 10th frame?
e.g. { first_roll: 5, second_roll: 5, third_roll: 3 }

lib/user_interface.rb
```diff
@@ -0,0 +1,85 @@
...
+class UserInterface
+  def initialize(io)
+    @io = io
+    @game = Game.new
+  end
+
+  def start_game
+    @io.puts "Welcome to your bowling scorecard!"
+    @io.puts "The game has started."
+    loop do
```
ZacMossHK, line 12: could be ```until @game.frames.length == 10```
and then could remove line 13

lib/game.rb
```diff
@@ -0,0 +1,34 @@
+require_relative './frame'
+require_relative './tenth_frame'
+
+class Game
+  attr_reader :frames
```
stephenfletchtek, line 5: Maybe just expose this by having a method?

lib/user_interface.rb
stephenfletchtek, line 41: method is too long (would fail rubcop). Suggest you refactor

PR #312, opened 2022-08-22
My bowling challenge effort

stephenfletchtek:


spec/frame_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'frame'
+
+RSpec.describe Frame do
+  context "frame contents" do
+    it "scores a frame" do
+      frame = Frame.new([4, 5])
```
naomischlosser, line 6: Aha, that's how you take one argument, but with 1 or 2 variables as needed!

lib/frame.rb
```diff
@@ -0,0 +1,32 @@
...
+    @frame = frame
+  end
+
+  def score
+    @frame.sum
+  end
+
+  def to_a
+    @frame
+  end
```
naomischlosser, line 15: If you have a couple of these, you might want to use attr_accessor (or reader) to shorten this bit of code.

lib/frame.rb
```diff
@@ -0,0 +1,32 @@
...
+    return true if @frame.sum == 10
+    return true if @frame.length == 3 && @frame[0..1].sum == 10
+    false
+  end
+
+  def strike
+    return true if spare_or_strike && @frame.length == 1
+    return true if @frame.length == 3 && @frame[0] == 10
+    false
+  end
```
naomischlosser, line 31: This is not that easy to follow. What is your reason to have spare_or_strike and strike instead of strike and spare only? Perhaps add a comment in your line of code to clarify.

lib/game.rb
```diff
@@ -0,0 +1,20 @@
+class Game
```
naomischlosser, line 1: We talked about this briefly this morning, you making two attempts at this bowling game, but I can't quite follow what the structure is (what file works on his own and which files belong together). Perhaps you can include this in a README file?

spec/game_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'game'
+
+RSpec.describe Game do
+  context "frame contents" do
+    it "no strikes" do
+      frames = 10.times.map { [3, 4] }.flatten
+      game = Game.new(frames.flatten)
```
naomischlosser, line 7: Do you need to flatten twice?

spec/game_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect(game.frames[0].score).to eq(7)
```
naomischlosser, line 8: Isn't this a test already being done in frame_spec.rb?

PR #313, opened 2022-08-30
Creating Pull request for code review

Curtis-Turk:


Curtis-Turk:
Adding my pull request for review

JavaScript/io.js
```diff
@@ -0,0 +1,75 @@
...
+  enterFrames() {
+    for (let f = 0; f < 10; f++) {
+      console.log(`Frame ${f + 1}`);
+      console.log("---------------");
+      console.log("frame: " + this.frames[f]);
+      this.promptRolls();
+      if (
+        f === 9 &&
+        (this.frames[f][0] + this.frames[f][1] === 10 ||
+          this.frames[f][1] === 10)
```
naomischlosser, line 26: As discussed, this will benefit from using the isStrike and isSpare function in the Score class so that if someone reads your code, it is a bit clearer.

JavaScript/io.js
```diff
@@ -0,0 +1,75 @@
+const prompt = require("prompt-sync")({ sigint: true });
```
naomischlosser, line 1: It doesn't really matter now, but in the future if the person who wants to use your code needs to install something, try to add it to the README.

JavaScript/io.js
```diff
@@ -0,0 +1,75 @@
...
+      }
+      console.log("---------------");
+      this.shoutScore(f);
+      console.log(" ");
+    }
+  }
+
+  promptRolls() {
+    let roll1 = prompt("First roll: ");
+    if (parseInt(roll1) > 10) {
```
naomischlosser, line 38: I used Number(roll1) for this, cool to know there is something else that can be used!

JavaScript/io.js
```diff
@@ -0,0 +1,75 @@
+const prompt = require("prompt-sync")({ sigint: true });
+const Score = require("./score");
+
+class Io {
```
naomischlosser, line 4: What do you think about this class name?

PR #316, opened 2022-09-26
Counter is working and works in terminal but no bonus or score

Vanboylan:


bowling_score.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def play
+    until @frame_count == 11
+        puts "frame: #{@frame_count}"
+        if @frame_count < 10
+            @frame = []
+            puts "how many pins?"
+            answer = gets.chomp.to_i
+            @frame << answer if answer >= 0 && answer <= 10
```
jonas-diete, line 17: Nice way of checking. What happens if the user enters a number outside these specifications or a string?

bowling_score.rb
```diff
@@ -0,0 +1,70 @@
...
+            if @frame.sum == 10
+                puts "strike!"
+                @frame == [10, 0]
+                @score << @frame
+                @card << 2
+            else
+                puts "how many pins?"
+                answer = gets.chomp.to_i
+                @frame << answer if answer + @frame[0] <= 10
```
jonas-diete, line 26: I like that you are checking whether the overall score is 10 or lower

bowling_score.rb
```diff
@@ -0,0 +1,70 @@
...
+        if @frame_count < 10
+            @frame = []
+            puts "how many pins?"
+            answer = gets.chomp.to_i
+            @frame << answer if answer >= 0 && answer <= 10
+            if @frame.sum == 10
+                puts "strike!"
+                @frame == [10, 0]
+                @score << @frame
+                @card << 2
```
jonas-diete, line 22: it's a bit hard to understand what @card does. It looks like it's recording whether a frame has a strike or a spare. Is there a way to give it a better name?

bowling_score.rb
```diff
@@ -0,0 +1,70 @@
...
+            answer = gets.chomp.to_i
+            @frame << answer if answer >= 0 && answer <= 10
+            puts "strike!" if @frame.sum == 20
+            puts "spare!" if @frame.sum == 10
+            if 
+            puts "how many pins?"
+            answer = gets.chomp.to_i
+            @frame << answer if answer >= 0 && answer <= 10
+            puts "strike!" if @frame.sum == 30
+            puts "spare!" if @frame.sum == 10
```
jonas-diete, line 49: Some of this could be a separate method/loop, as you are getting similar user input every time

jonas-diete (commented):
Well done, Jase. Looks like you have got an MVP and it can score simple games without bonuses. I think it would be worth adding some tests to make refactoring and continued working on it easier and quicker.

PR #320, opened 2022-10-16
Bowling Challenge Working Logic Implemented

kateusacova:
- Full game logic is implemented and tested, appears to be working as expected.
- Minimal UI.
- No user input validation.
- Probably need to do more refactoring and split into sub-classes.


design.md
```diff
@@ -0,0 +1,73 @@
+App flow:
+- 1: welcome_message: Welcome to Bowling Scorecard Program!
+- 2: instructions: Please enter your scores as prompted to receive your final score in the end. 
+
+
```
tomallens, line 5: Super concise plan with a well-defined idea of how to progress, very impressive.

Even more so if the logic was apparent before writing tests/sketching methods.

lib/app.rb
```diff
@@ -0,0 +1,81 @@
+class Application
+
+  attr_accessor :frames
+
+  def initialize(io)
+    @io = io
+    @frames = [[],[],[],[],[],[],[],[],[],[]]
```
tomallens, line 7: Probably the most concise storage option for game state possible, agree with this implementation!

lib/app.rb
```diff
@@ -0,0 +1,81 @@
...
+    if previous_frame[0] == 10
+      frame.sum
+    elsif previous_frame[0..1].sum == 10
+      frame[0]
+    else 
+      0
+    end
+  end
+
+  def add_user_scores(frame)
```
tomallens, line 54: Love this method! Would have been very easy to get trapped putting this logic elsewhere but this both keeps it super readable and eliminates a lot of space for repetition.

lib/app.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  private
+  
+  def third_roll
+    @io.puts "Roll 3:"
+    @frames[9] << get_user_input
+  end
+
+  if __FILE__ == $0
```
tomallens, line 77: Have to be honest I'm not sure what's going on here!

Is this just covering in case of future save/load functionality?

If so, super smart but technically should just be `run` here for this build.
If not, congratulations for outsmarting me!

lib/app.rb
```diff
@@ -0,0 +1,81 @@
...
+    play_first_nine_frames
+    play_final_frame
+    @io.puts "Your final score is: #{@frames.flatten.sum}"
+  end
+
+  def play_first_nine_frames
+    @frames[0..8].each_with_index do |frame, i|
+      @io.puts "Frame #{i+1}:"
+
+      frame == add_user_scores(frame)
```
tomallens, line 20: Should this be a logical operator or a single `=` ? I'm not sure if this line is best practice or not but either way it's confusing to read for me. Hopefully this is a reviewer quality issue and not code quality!

lib/app.rb
```diff
@@ -0,0 +1,81 @@
...
+  def get_user_input
+    user_score = @io.gets.chomp.to_i
+  end
+
+  private
+  
+  def third_roll
+    @io.puts "Roll 3:"
+    @frames[9] << get_user_input
+  end
```
tomallens, line 75: The final frames are a big ball of logic to work through but I think I agree with your implementation here - splitting functions much further is just going to compromise on readability.

spec/app_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+      result = [[10],[2,3,5],[2,3,0],[3,7,0],[2,3,2],[10,0],[10,10],[2,8,10],[10,10],[2,7]]
+
+      app.run
+
+      expect(app.frames).to eq result
+    end
+  end
+
+end
+
```
tomallens, line 113: potentially could have made shorthand methods for testing e.g 

```
def puts?(text)
  expect(io).to receive(:puts).with(text)
end
```
but that's just style points I think!

tomallens: Overall this looks like a solid testing regimen - I'd love to know what kind of test coverage you're getting - perhaps something for a readme?

tomallens (commented):
Incredibly elegant solution.

The app could potentially have been split into classes to make a better foundation for further development but given the simplicity of the specification, the single file feels like a much more rational delivery given the time available.

What looks like a solid battery of tests, let down only slightly by a lack of self-promotion showing them off! Adding a readme with an example output, showing the 'nice-to-haves' as your direction for further iteration, and showing off test coverage etc. (because it looks like it would be great already!) would make it easier to appreciate how good the code is to people browsing your repos.

Really inspiring stuff 🥇

lib/app.rb
kateusacova: If I only had this:
 `app = Application.new(Kernel)
  app.run`
Every time I run RSpec, it would try to run the app. Because I have this ` if __FILE__ == $0`,  I can run the app from the terminal and I can also run my tests without wasting time deleting `app.run`. Hope I explained it well?

lib/app.rb
kateusacova: You're right, it should have been just a single =. Since the code still works, I don't think I need this line at all, I will remove it :)

spec/app_spec.rb
kateusacova: Yeah, I didn't install simplecov to see the coverage but I think it should be around 95-100% :)

lib/app.rb
kateusacova: No, I def agree. I should have split it further into self - explanatory methods. Hopefully, will do it this weekend :)

PR #325, opened 2022-10-17
incomplete attempt

HibaqObsiye:


lib/frame.rb
```diff
@@ -0,0 +1,22 @@
+require './lib/score_card.rb'
+class Frame
+attr_accessor :roll_one, :roll_two
```
eggs-benny, line 3: Nice use of attr_accessor - definitely makes the code neater!

lib/frame.rb
```diff
@@ -0,0 +1,22 @@
...
+        @roll_one == 10
+    end
+
+    def sum
+        @roll_two + @roll_one
+    end
+
+    def spare?
+        @roll_one != 10 && sum == 10
+    end
```
eggs-benny, line 20: Nice and simple logic here.

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+
+         @frames[number].sum + 10
+
+        elsif @frames[number - 1].spare?
+            @frames[number].roll_one + 10
+        else
+            @frames[number - 1].sum
+        end
+    end
```
eggs-benny, line 25: You're nearly there with the logic by the looks of it. But I think having some RSpec tests will help you tell fairly quickly what's working and what's not.

It's also worth considering whether to keep track of the number of frames, so that you can add in the logic for if you get a strike or spare in the 10th frame.

spec/frame_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    context "players knocks down 10 pins in two rolls " do
+        it "returns true for spates " do
+            frame = Frame.new(5, 5)
+            expect(frame.spare?).to eq(true)
+            expect(frame.strike?).to eq(false)
+            expect(frame.sum).to eq(10)
+        end
+    end
```
eggs-benny, line 29: All solid tests!

eggs-benny (commented):
Overall it's a solid base, just keep going and you should get there fairly soon!

PR #326, opened 2022-10-17
nothing to push

tomallens:
I have a pretty picture and that's it


kateusacova (commented):
I think you planned your application well. You thought about all the details necessary for the app to work well (bonus, table to display score in user-friendly format, the data structure for your score tracker). I don't have anything to add to your planning :)

PR #331, opened 2022-10-17
Bowling challenge - Na-Rae

Nameo91:
Counted score for strike and spare

but need some refactoring


lib/score_card.rb
```diff
@@ -14,7 +14,13 @@ def all
   def total_score
     score = 0
     @frames.each_with_index do |frame, i|
-      if frame.spare? && frame != last_bonus_game
+      if frame.strike? && frame != last_bonus_game
+        if @frames[i + 1].strike? && @frames[i + 1] != last_bonus_game
```
laurenbrabbin, line 18: could look at putting this if/else statement into a seperate method

spec/score_card_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      score_card.add(frame)
+      score_card.add(frame)
+      score_card.add(frame)
+      score_card.add(frame)
+      score_card.add(Frame.new(10, 10))
+      score_card.add(Frame.new(10, 0))
+
+      expect(score_card.total_score).to eq(300)
+    end
+  end
```
laurenbrabbin, line 103: lots of tests which is great to see

PR #333, opened 2022-10-17
Joe's Bowling Challenge

JoeOsborne77:
Incomplete. Still needs;

- Spare and strike functions properly working
- More tests 
- IO / doubles properly implemented


application.rb
```diff
@@ -0,0 +1,57 @@
...
+      if @roll_1 < 10
+        @scorecard << @roll_1
+        second_throw
+      else
+        @scorecard << @roll_1        
+        @round += 1 
+      end
+    end
+  end
+
```
Cragg87, line 26: It would be useful to have some comments so that a reviewer can quickly identify what a method does.

application.rb
```diff
@@ -0,0 +1,57 @@
...
+      strike_calculator
+      puts "Thanks for playing. Your total score is #{@scorecard.sum}"
+      break
+    end
+    
+    puts "Please enter the score from your first roll from @round #{@round}"  
+    @roll_1 = gets.chomp.to_i
+      if @roll_1 < 10
+        @scorecard << @roll_1
+        second_throw
```
Cragg87, line 19: This is good - I wasn't sure how to run another method from within a method.

application.rb
```diff
@@ -0,0 +1,57 @@
...
+    @roll_2 = gets.chomp.to_i
+    @round += 1 
+    if @roll_1 + @roll_2 == 10
+      @scorecard << @roll_2
+      spare
+      puts "Tally so far is #{@scorecard}"  
+    end
+    first_throw
+  end
+
```
Cragg87, line 43: Could 'spare' and 'strike' become booleans which, when true, affect the next frame, before being reset to false?

application.rb
```diff
@@ -0,0 +1,57 @@
+class Application
```
Cragg87, line 1: Maybe include a design recipe for transparency?

application.rb
```diff
@@ -0,0 +1,57 @@
...
+  def initialize(scorecard, round)
```
Cragg87, line 2: This is good - I wasn't sure how to run another method from within a method.

application.rb
Cragg87, line 26: It would be useful to have some comments so that a reviewer can quickly identify what a method does.

application.rb
```diff
@@ -0,0 +1,57 @@
...
+    @round += 1 
+    if @roll_1 + @roll_2 == 10
+      @scorecard << @roll_2
+      spare
+      puts "Tally so far is #{@scorecard}"  
+    end
+    first_throw
+  end
+
+  def spare
```
Cragg87, line 44: Could 'spare' and 'strike' become booleans which, when true, affect the next frame, before being reset to false?

Cragg87 (commented):
Great work so far!

PR #372, opened 2023-02-20
Bowling challenge- Ruby

ChalkandFeather:
Tests passed, for strike, spare and frame count. Requires further testing to ensure functionality.


lib/frame_repository.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+  
+  def add(roll)
+    while true do
+      #break if roll.index[0] == 10 
+      #return @frame_score << roll.index[0] + 0
+      #else
+      @frame_score << roll
+      break if @frame_score.length <= 2
+   end
```
JohnForster, line 15: I don't think this loop is necessary. The first and second time we add a roll, the `@frame_score` array will contain one or two elements and so we break out of the loop immediately. If we try adding a third element to this array, the condition on L14 will evaluate to false, and we will loop infinitely, adding `roll` to `@frame_score` indefinitely.

lib/frame_repository.rb
```diff
@@ -0,0 +1,21 @@
+#require 'frame'
+
+class FrameRepository
```
JohnForster, line 3: The term `Repository` typically indicates storage of some kind, and is used when writing classes that interact with a database. `Frame` is a good name for this class, as each _instance_ represents one frame of the game.

lib/Game_repository.rb
```diff
@@ -0,0 +1,31 @@
...
+      elsif @rolls[roll_index] +@rolls[roll_index +1] != 10 # frame score accumulation <10
+      score += @rolls[roll_index] + @rolls[roll_index + 1] 
+      elsif @rolls[roll_index] == 10 # Check Strike- previous  fail appeared to be due to rspec - further tests needed to ensure- bonus score is calculated the bonus feature.
+      score += @rolls[roll_index] + @rolls[roll_index + 1] + @rolls[roll_index + 2]
+      roll_index += 1
+   end
+      roll_index += 2
+     end
+      score
+    end
```
JohnForster: This could be refactored to improve readability by creating descriptive variable names and extracting some logic into separate functions. For example:

```rb
def score_card
  roll_index = 0
  score = 0
  
  10.times do
    roll = @rolls[roll_index]
    next_roll = @rolls[roll_index + 1]
    if spare?(roll, next_roll)
      subsequent_roll = @rolls[roll_index + 2]
      score += roll + next_roll + subsequent_roll
    elsif !spare(roll, next_roll)
      # etc...
    end
  end
end
```

JohnForster: Additionally, because the first `if` and `elsif` statements cover all possible outcomes (Do the current roll and the next roll add up to ten?) the later `elsif` will never be run.

JohnForster (commented):
You've tackled a tricky logical problem here and come up with a good solution to it, but your code structure means that it can be difficult to understand and change. Your `FrameRepository` class was a good first step, but the second step would be to use it inside your `GameRepository` class.

A game of bowling consists of a series of frames, so your `GameRepository` class should probably hold a list of these frames (in an array), then receive rolls, and decide which frame to add the roll to. When scoring up the whole game, the `GameRepository` class just needs to ask each frame for its score, and add them up.

Your testing is clear and thorough, and demonstrates a good TDD process 👍 

Please let me know if you have any questions about this feedback

PR #388, opened 2023-03-20
Bowling Challenge

leah-massey:
Screenshot 2023-03-20 at 10 13 47


jpates1:
Overall well thought through design and logic. Loved the sequence diagram, clear you've spent a lot of time thinking it all through. 

- Good use of names and naming conventions for classes, variables and methods. 
- Code is fairly easy to read and follow, although you could consider adding some comments before methods to explain their function to the reader (when you're finished! Of course). 
- There is probably some scope to refactor some of the methods, and make them private to clean things up. 
- Testing so far is thorough and thoughtful. 

Think its going to be great when you're done!

PR #390, opened 2023-04-21
bowling scorecard challenge

umutbaykan:
validating whether a user is allowed to get additional shots (after strike/sparing the 10th one) is assumed to be handled by the UI so havent been incorporated


lib/bowl.rb
```diff
@@ -0,0 +1,71 @@
...
+    message = ""
+    if frame[0] == 10
+      message = "strike"
+    elsif frame.sum == 10
+      message = "spare"
+    end
+    return message
+  end
+
+  def add_bonus_frame(*args)
```
SomthingInteresting, line 27: Based on the focus points provided to us I would suggest encapsulating the Frame logic into a seperate class.  This would make it more modular and the bowlingscorer class would be only responsible for managing the Frames rather than the logic.

lib/bowl.rb
```diff
@@ -0,0 +1,71 @@
...
+      @player_score += count_frame_score(i)
+      i += 1
+    end
+    count_player_bonus_scores if @frames.length > 10
+    return @player_score
+  end
+
+  ## Helper methods - Used in testing and not relevant to gameplay ##
+  def frames
+    @frames
```
SomthingInteresting, line 64: I assume if you had multiple classes you would've used accessor methods to access these instance variables instead?

spec/bowl_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+    end
+
+    it "fails if user inputs negative numbers" do
+      expect{bowl.add_frame(-1,5)}.to raise_error "Smells like invalid input"
+    end
+
+    it "fails if user inputs negative numbers" do
+      expect{bowl.add_frame(4,-5)}.to raise_error "Smells like invalid input"
+    end
+
```
SomthingInteresting, line 36: Good fail scenarios but would suggest one for non numeric input, floating point and maybe other types but that's just me nitpicking.

lib/bowl.rb
```diff
@@ -0,0 +1,71 @@
...
+    @frames << [*args]
+    return [*args]
+  end
+
+  def count_player_bonus_scores
+    fail "Player has not scored any bonuses!" if @frames.length <= 10
+    @player_score += @frames[-1].flatten.sum
+  end
+
+  def count_frame_score(index)
```
SomthingInteresting, line 37: Would be good to split the responsibilities out here as it currently calculates the frame score and the special cases.

SomthingInteresting (commented):
The code has a bunch of tests that cover both state and behavior, making it pretty solid. It's easy to read and organised, plus it's got some nice error handling with clear explanations.

On the flip side, it could do a better job sticking to the Single Responsibility Principle since some methods are juggling a few tasks at once. Splitting them up would make the code even better.

Overall, it's a good piece of code, but with a few tweaks, it could be even better!

PR #391, opened 2023-04-23
Bowling challenge in Ruby

tomcarmichael:


lib/frame.rb
bookloversolmaz, line 1: The frame class looks good, you've considered first and second rolls and how they interact with bonus rolls

lib/game.rb
bookloversolmaz, line 1: This looks good, comments are helpful and you've included methods for calculating bonus points in frams 1 to 9 and in the final frame.

spec/game_spec.rb
bookloversolmaz, line 1: Good idea using the example game from Github in the test.

spec/frame_spec.rb
bookloversolmaz, line 1: The only suggestion I can make is to add more tests here.

PR #394, opened 2023-04-24
Partially completed bowling challenge

SomthingInteresting:
Need to work on bonus point implementation.


README.md
```diff
@@ -1,65 +1,39 @@
...
+This is a solution to the bowling challenge in Ruby. The challenge is to write a program that calculates the score of a bowling game.
 
-## The Task
+## How to use
 
-**THIS IS NOT A BOWLING GAME, IT IS A BOWLING SCORECARD PROGRAM. DO NOT GENERATE RANDOM ROLLS. THE USER INPUTS THE ROLLS.**
+Clone this repo and run bundle install to install the dependencies.
 
-Count and sum the scores of a bowling game for one player. For this challenge, you do _not_ need to build a web app with a UI, instead, just focus on the logic for bowling (you also don't need a database). Next end-of-unit challenge, you will have the chance to translate the logic to Javascript and build a user interface.
+To run the tests, run rspec from the command line.
```
umutbaykan, line 9: Currently the test on scorecard_spec.rb line 38 fails on rspec, but I am assuming that is because it is still in WIP

README.md
```diff
@@ -1,65 +1,39 @@
...
-* Commit often, with good commit messages
-* Single Responsibility Principle and encapsulation
-* Clear and readable code
+## Classes
 
-## Bowling — how does it work?
+### Game
 
-### Strikes
+The main class that controls the flow of the game. It has an `initialize` method that sets up the game with a new Frame and a new Scorecard. It also has a `roll` method which calls the roll method on the current frame and updates the Scorecard. The game continues until all frames are played.
```
umutbaykan, line 27: I get the logic, but do we really need a Roll class? It seems like all it is doing right now is keeping an integer for a score, could we have not done that by just using actual integers in an array?

lib/game.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def roll(pins)
+    roll = Roll.new(pins)
+    @current_frame.rolls << roll
+    @scorecard.update(@current_frame, @frames)
+    update_frame
+  end
+
+  def update_frame
+    @roll_count += 1
```
umutbaykan, line 23: does this need to be a method seeing as it is a one-line code which is only used once?

lib/roll.rb
```diff
@@ -0,0 +1,7 @@
+class Roll
+  attr_reader :score
+
+  def initialize(score)
+    @score = score
+  end
+end
```
umutbaykan, line 7: see the comment on readme, I think splitting the rolls into a different class over-complicates it

lib/game.rb
```diff
@@ -0,0 +1,30 @@
...
+require_relative 'roll'
+
+class Game
+  attr_reader :scorecard, :current_frame, :frames, :roll_count
+
+  def initialize
+    @scorecard = Scorecard.new
+    @current_frame = Frame.new
+    @frames = [@current_frame]
+    @roll_count = 0
```
umutbaykan, line 12: I like the logic of splitting the scorecard in a different class, but it seems like you need to update the frames in scorecard class as well as your current game class, effectively duplicating the information. Can you not have a single array of frames record what the player has scored and all the classes could apply their logic on that one?

lib/scorecard.rb
```diff
@@ -0,0 +1,32 @@
...
+  def initialize
+    @frames = []
+  end
+
+  def update(frame, frames)
+    frame.total = frame.rolls.sum(&:score)
+    if frame.rolls.size == 2 || frame.rolls.map(&:score).sum >= 10
+      @frames << frame
+    end
+  end
```
umutbaykan, line 13: see line 25 on game.rb, which is very similar. I think scorecard and game classes are slightly creeping into each other's territory, resulting in duplicating information / code

lib/scorecard.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  private
+
+  def format_frame(index, frame, score)
+    frame_info = "FRAME #{index + 1}:"
+    rolls_info = frame.rolls.each_with_index.map { |roll, r_index| "ROLL #{r_index + 1}: #{roll.score}" }.join(' ')
+    "#{frame_info} #{rolls_info} | TOTAL: #{frame.total} | SCORE: #{score}\n"
+  end
+end
```
umutbaykan, line 32: I like this portion of this class. It is good practice to keep this strictly to printing / displaying scores of the user

spec/scorecard_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      # Roll a strike
+      game.roll(10)
+
+      # Roll two more rolls
+      game.roll(3)
+      game.roll(4)
+
+      expected_scorecard = "FRAME 1: ROLL 1: 10 | TOTAL: 17 | SCORE: 17\n" \
+                           "FRAME 2: ROLL 1: 3 ROLL 2: 4 | TOTAL: 7 | SCORE: 24\n"
```
umutbaykan, line 36: this test currently does not pass, but I am assuming you will implement strikes / spares next

umutbaykan (commented):
I think a good start!
A few general comments:
Moving forward, I think there needs to be a defined line between game and scorecard class, as in which one is adding the rolls into the frames and which one is interpreting those scores and including spares / strikes where relevant. Currently as the player takes shots, the game copies that information between multiple classes. Instead there could be a singular class / source which holds all the records of the player and each different class could apply their methods using the same source. This would help reduce the complexity when introducing methods to calculate strike / spare scores.

Current tests seem to be covering the basics and need to be expanded to cover edge cases and actual game simulations, but the project is still in development so I am sure you will add more as you go. 

If you are incorporating UI, I would also suggest adding a few validators to check if the user enters invalid input (string inputs, negative inputs etc.).

spec/scorecard_spec.rb
SomthingInteresting: Yes that's right.  Started test driving and then branched off.

lib/scorecard.rb
SomthingInteresting: Agreed.  I will take the time to refactor the classes and ensure the functions are seperate.

lib/roll.rb
SomthingInteresting: It came about only from trying to make the program modular.  Agreed it doesn't really suit this program though.

README.md
SomthingInteresting: Correct WIP

PR #397, opened 2023-04-24
bowling challenge

oanapintoiu:


lib/bowling_game.rb
```diff
@@ -0,0 +1,80 @@
...
+    @rolls = []
+  end
+
+
+  def roll(pins)
+    @rolls.push pins
+  end
+
+
+  def score
```
toppy007, line 12: I really like this code very simple and does a job.

toppy007 (commented):
very good program!

PR #398, opened 2023-04-24
Solmaz's partially completed bowling challenge

bookloversolmaz:


spec/score_spec.rb
```diff
@@ -0,0 +1,36 @@
+require "score"
+
+RSpec.describe Score do
+  # The score should be 0 before the player starts
+  it 'initializes' do
+    score = Score.new
+    expect(score.score_count(0)).to eq(0)
+  end
+
+  context "players score less than 10 in single frame" do
```
tomcarmichael, line 10: nice use of context to make the test readable as documentation

spec/score_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it "adds the score 3 to counter which is at 0" do
+      score = Score.new
+      score.score_count(3)
```
tomcarmichael, line 13: perhaps line 13 is not necessary here

spec/score_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      expect(score.score_count(3)).to eq (3)
+    end
+    it "first roll knocks down 5, second 3" do
+      score = Score.new
+      score.score_count(5)
+      score.score_count(3)
+      expect(score.score_count(8)).to eq (8)
```
tomcarmichael, line 20: could perhaps create a new method with no args to just read the score for testing as this may be a self fulfilling test at the moment

tomcarmichael (commented):
Looks like a really good start and I i like the approach to calculating strikes and spares with FX notation

PR #399, opened 2023-04-24
Fixed forking issue

aryadath:


lib/scoreboard.rb
```diff
@@ -0,0 +1,34 @@
+class Scoreboard
+  def initialize
```
ManuelaIacobovici, line 2: Maybe we should initialise with a collection of rolls.

lib/scoreboard.rb
```diff
@@ -0,0 +1,34 @@
...
+    @rolls = [] # array that will store down number of knocked pins after each roll
+    @current_roll = 0 #to keep track of the current roll. Initalising means it is ready to take information so it 
+    # can calculate the score
+  end
+
+  def roll(pins)
```
ManuelaIacobovici, line 8: We are not calling this function anywhere in the code.

lib/scoreboard.rb
```diff
@@ -0,0 +1,34 @@
+class Scoreboard
+  def initialize
+    @rolls = [] # array that will store down number of knocked pins after each roll
+    @current_roll = 0 #to keep track of the current roll. Initalising means it is ready to take information so it
```
ManuelaIacobovici, line 4: This could be maybe a reference to the frame, once we detect that we should move to the next frame. It is not used in the code.

ManuelaIacobovici (commented):
Very well written code overall. Great job!

